PR #3, opened 2022-10-04
Allow switching between release channels

GeckoEidechse:
Right now it's a single toggle button. In the future it should be a dropdown menu

### DIFF
--- a/src-vue/src/plugins/store.ts
+++ b/src-vue/src/plugins/store.ts
@@ -184,7 +184,7 @@ function _initializeListeners(state: any) {
  * This retrieves Northstar version tag, and stores it in application
  * state, for it to be displayed in UI.
  */
-async function _get_northstar_version_number(state: any) {
+export async function _get_northstar_version_number(state: any) {
     let northstar_version_number: string = await invoke("get_northstar_version_number_caller", { gamePath: state.game_path });
     if (northstar_version_number && northstar_version_number.length > 0) {
         state.installed_northstar_version = northstar_version_number;
--- a/src-vue/src/plugins/store.ts
+++ b/src-vue/src/plugins/store.ts
@@ -184,7 +184,7 @@ function _initializeListeners(state: any) {
  * This retrieves Northstar version tag, and stores it in application
  * state, for it to be displayed in UI.
  */
-export async function _get_northstar_version_number(state: any) {
+async function _get_northstar_version_number(state: any) {
     let northstar_version_number: string = await invoke("get_northstar_version_number_caller", { gamePath: state.game_path });
     if (northstar_version_number && northstar_version_number.length > 0) {
         state.installed_northstar_version = northstar_version_number;
--- a/src-vue/src/views/DeveloperView.vue
+++ b/src-vue/src/views/DeveloperView.vue
@@ -53,7 +53,7 @@ export default defineComponent({
             ElNotification({
                 title: `${this.$store.state.release_canal}`,
                 message: `Switched release channel to: "${this.$store.state.release_canal}"`,
-                type: 'error',
+                type: 'success',
                 position: 'bottom-right'
             });
         }
--- a/src-vue/src/views/DeveloperView.vue
+++ b/src-vue/src/views/DeveloperView.vue
@@ -19,7 +19,6 @@ import {defineComponent} from "vue";
 import { invoke } from "@tauri-apps/api";
 import { ElNotification } from "element-plus";
 import { ReleaseCanal } from "../utils/ReleaseCanal";
-import { _get_northstar_version_number } from "../plugins/store"
 
 export default defineComponent({
     name: "DeveloperView",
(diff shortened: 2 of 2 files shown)


src-vue/src/plugins/store.ts
```diff
@@ -184,7 +184,7 @@ function _initializeListeners(state: any) {
  * This retrieves Northstar version tag, and stores it in application
  * state, for it to be displayed in UI.
  */
-async function _get_northstar_version_number(state: any) {
+export async function _get_northstar_version_number(state: any) {
```
Alystrasz: This method should stay private (not get exported), and should be invoked through the associated store mutation.
Check https://github.com/GeckoEidechse/FlightCore/blob/main/src-vue/src/views/SettingsView.vue#L50 as an example.

src-vue/src/views/DeveloperView.vue
```diff
@@ -29,6 +35,26 @@ export default defineComponent({
                 type: 'error',
                 position: 'bottom-right'
             });
+        },
+        async toggleReleaseCandidate() {
+            // Flip between RELEASE and RELEASE_CANDIDATE
+            if (this.$store.state.release_canal === ReleaseCanal.RELEASE) {
```
Alystrasz, line 40: This can be done with a [ternary operator](https://www.c-sharpcorner.com/UploadFile/5089e0/ternary-conditional-operator-in-typescript/) to be more readable.

```typescript
this.$store.state.release_canal = this.$store.state.release_canal === ReleaseCanal.RELEASE
    ? ReleaseCanal.RELEASE_CANDIDATE
    : ReleaseCanal.RELEASE;
```

src-vue/src/views/DeveloperView.vue
```diff
@@ -29,6 +34,28 @@ export default defineComponent({
...
+                this.$store.state.release_canal = ReleaseCanal.RELEASE_CANDIDATE;
+            }
+            else {
+                this.$store.state.release_canal = ReleaseCanal.RELEASE;
+            }
+
+            // Update current state so that update check etc can be performed
+            this.$store.commit("initialize");
```
Alystrasz, line 48: Shouldn't you be calling `checkNorthstarUpdates` instead of `initialize`?
`initialize` should be only called on application boot.

src-vue/src/views/DeveloperView.vue
```diff
@@ -29,6 +34,28 @@ export default defineComponent({
...
+
+            console.log(this.$store.state)
+
+            // Display notification to highlight change
+            ElNotification({
+                title: `${this.$store.state.release_canal}`,
+                message: `Switched release channel to: "${this.$store.state.release_canal}"`,
+                type: 'error',
```
Alystrasz, line 56: I don't think this is an error message; you should use [success or info type](https://element-plus.org/en-US/component/notification.html#with-types) instead.

Alystrasz (changes requested):

src-vue/src/views/DeveloperView.vue
Alystrasz: For code to be even more readable, we could put `release_canal` in the computed properties of the DeveloperView component and [add associated getter/setter](https://codingexplained.com/coding/front-end/vue-js/adding-getters-setters-computed-properties):

```typescript
computed: {
	release_canal: {
		get: function(): string {
      			return this.$store.state.release_canal;
		},
		set: function(newValue: string) {
			this.$store.state.release_canal = NorthstarReleaseType[newValue];
		}
	}
}
```

Which would enable writing:

```typescript
this.release_canal = this.release_canal === ReleaseCanal.RELEASE
    ? ReleaseCanal.RELEASE_CANDIDATE
    : ReleaseCanal.RELEASE;
```

This is up to you ;)

src-vue/src/views/DeveloperView.vue
GeckoEidechse: Whoops forgot to update from testing, fixed in 0085a57a3d35c340552859a8d8ae95df6ba5e8b8

src-vue/src/views/DeveloperView.vue
GeckoEidechse: Fixed in 3398c073eb57c4309770a82b1193c198744d7ad6

src-vue/src/views/DeveloperView.vue
GeckoEidechse: I went with just the ternary operator in 05c945aaa13bbb20859ae4eb4eff98a0a1e91211 for the sole reason that simply copy-pasting the above snippet did not work as `NorthstarReleaseType` is not recognised :P

Alystrasz:
Eh, I didn't even validated this PR :open_mouth:

GeckoEidechse:
> Eh, I didn't even validated this PR :open_mouth:

Gotta move fast to crank out new features >:)

PR #5, opened 2022-10-05
chore: Remove unused code and dependencies

GeckoEidechse:
Removes unused `index.html` and corresponding styling which is a leftover from before the UI rewrite.

Also removes the unused npm dependencies that are leftover from when the frontend packages where defined in the root directory.


Alystrasz (changes requested):
Looks overall good.
We should however remove `dist` folder from git tracking with a `.gitignore` entry,

GeckoEidechse:
> Looks overall good. We should however remove `dist` folder from git tracking with a `.gitignore` entry,

Do we need to? Cause nothing is generating that folder anymore with the removal of these last few remaining files ^^

Alystrasz:
Well, we don't need this, and I think it's created while release building
I'll let you give it a try :)

GeckoEidechse:
With a549c4e87b4d9dada8cd63568833cb8ff77ea7d5, the resulting vue build should stay in `src-vue/dist` which is properly gitignore so it should be fine. ^^

Imma merge and give it a try :P

PR #6, opened 2022-10-05
feat: Initial logic to disable all but core mods

GeckoEidechse:
In the future, this button should be moved to a "Repair" view. For now I'm putting it into Dev view for testing.

### DIFF
--- a/src-vue/src/views/DeveloperView.vue
+++ b/src-vue/src/views/DeveloperView.vue
@@ -78,7 +78,7 @@ export default defineComponent({
                 .catch((error) => {
                     ElNotification({
                         title: 'Error',
-                        message: `${error}`,
+                        message: error,
                         type: 'error',
                         position: 'bottom-right'
                     });
--- a/src-vue/src/views/DeveloperView.vue
+++ b/src-vue/src/views/DeveloperView.vue
@@ -68,7 +68,6 @@ export default defineComponent({
                 install_type: this.$store.state.install_type
             } as GameInstall;
             await invoke("disable_all_but_core_caller", { gameInstall: game_install }).then((message) => {
-                console.log(message);
                 ElNotification({
                     title: 'Success',
                     message: "Disabled all mods but core",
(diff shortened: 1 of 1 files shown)


src-vue/src/views/DeveloperView.vue
```diff
@@ -53,6 +61,30 @@ export default defineComponent({
...
                 position: 'bottom-right'
             });
+        },
+        async disableAllModsButCore() {
+            let game_install = {
+                game_path: this.$store.state.game_path,
+                install_type: this.$store.state.install_type
+            } as GameInstall;
+            await invoke("disable_all_but_core_caller", { gameInstall: game_install }).then((message) => {
+                console.log(message);
```
Alystrasz, line 71: This should be removed.

src-vue/src/views/DeveloperView.vue
```diff
@@ -53,6 +61,30 @@ export default defineComponent({
...
+                ElNotification({
+                    title: 'Success',
+                    message: "Disabled all mods but core",
+                    type: 'success',
+                    position: 'bottom-right'
+                });
+            })
+                .catch((error) => {
+                    console.error(error);
```
Alystrasz, line 80: This should be removed.

src-vue/src/views/DeveloperView.vue
```diff
@@ -53,6 +61,30 @@ export default defineComponent({
...
+                    ElNotification({
+                        title: 'Error',
+                        message: `${error}`,
```
Alystrasz, line 83: Why not `message: error`?

Alystrasz (changes requested):

src-vue/src/views/DeveloperView.vue
GeckoEidechse: Done with 3359b95eb3defa968ee391bc9ca7426754e5e78f

src-vue/src/views/DeveloperView.vue
GeckoEidechse: Done with 3359b95eb3defa968ee391bc9ca7426754e5e78f

PR #7, opened 2022-10-06
feat: Manually find game if needed

### LINKED ISSUE #4: When Titanfall2 install path not detected, use play/update/install button to select location
Currently it displays an `alert()` which is an issue of its own and then also doesn't allow you to select location using the install/update/play button (I will refer to it as omni button from now on).

A smoother UX would be to use an [Element Plus](https://element-plus.org/) notification and re-use the omni button to select install location.

Alystrasz:
Closes #4.

##### TODOs

- [x] Display a notification if path was chosen successfully
- [x] Remove permanent notification on successful path selection
- [x] Change button text ("Find game"?)
- [x] https://github.com/GeckoEidechse/FlightCore/pull/7#discussion_r990559835


Alystrasz:
I suppose that's because the secret you created on your repo is not shared to forks (for privacy reasons)

src-vue/src/plugins/store.ts
```diff
@@ -59,6 +63,47 @@ export const store = createStore<FlightCoreStore>({
...
+                alert("Please only select a single directory");
+            } else if (selected === null) {
+                // user cancelled the selection
+            } else {
+                // user selected a single directory
+
+                // Verify if valid Titanfall2 install location
+                let is_valid_titanfall2_install = await invoke("verify_install_location", { gamePath: selected }) as boolean;
+                if (is_valid_titanfall2_install) {
+                    state.game_path = selected;
```
GeckoEidechse, line 84: `state.install_type` needs to be also set (preferably to [`UNKNOWN`](https://github.com/GeckoEidechse/FlightCore/blob/52394a3e86373801f5285843d6931a692a619be1/src-vue/src/utils/InstallType.ts#L7)).

Otherwise this error message pops up when trying to launch the game after manually selecting the game path ^^

screenshot

Alystrasz: This shouldn't happen anymore.

src-vue/src/components/PlayButton.vue
```diff
@@ -13,7 +13,7 @@ export default defineComponent({
 
             switch(this.$store.state.northstar_state) {
                 case NorthstarState.GAME_NOT_FOUND:
-                    return "Titanfall2 not found";
+                    return "Select game folder";
```
GeckoEidechse, line 16:
```suggestion
                    return "Select Titanfall2 game folder";
```

Might be too long but would make it more obvious what the button does ^^

src-vue/src/views/SettingsView.vue
```diff
@@ -36,37 +32,7 @@ export default defineComponent({
...
-                else {
-                    // Not valid Titanfall2 install
-                    ElNotification({
-                        title: 'Wrong folder',
-                        message: "Selected folder is not a valid Titanfall2 install.",
-                        type: 'error',
-                        position: 'bottom-right'
-                    });
-                }
-            }
```
GeckoEidechse, line 69: Shouldn't the user still be able to update the path in the settings menu? ^^

Alystrasz: This code was actually moved in the `updateGamePath` method (https://github.com/GeckoEidechse/FlightCore/pull/7/files#diff-32bc04ae938110cf119ad14d3b3c167056516fea1f071eb1d45c1804dbc7ed4dR73), so it's used when:
* user clicks omnibutton when game folder was not found
* user clicks folder button in settings view

GeckoEidechse (approved):
Code looks fine and works as intended when testing on Linux.

PR #8, opened 2022-10-06
fix: Disable self-updater in CI builds from push

GeckoEidechse:
Remedies the issue that anyone pushing without access to private key will see their CI fail.


GeckoEidechse:
Was worried that `jq` might be missing on CI machines but it seems to pass.

@Alystrasz, maybe you wanna trigger a CI with this commit to make sure that it properly disabled self-update ^^

GeckoEidechse:
Seems to work as per https://github.com/GeckoEidechse/FlightCore/actions/runs/3207126073/jobs/5241684241 in #10

PR #11, opened 2022-10-07
feat: Show own version number in settings view

GeckoEidechse:
Adds a new field to state that stores FlightCore version number. Version number is pulled from backend on application launch.

version number showcase


src-vue/src/views/SettingsView.vue
```diff
@@ -26,6 +28,11 @@ import { ElNotification } from 'element-plus';
 
 export default defineComponent({
     name: "SettingsView",
+    computed: {
+        flightcoreVersion(): string {
+            return this.$store.state.flightcore_version; // TODO change to FlightCore version
```
Alystrasz, line 33: Still TODO? :)

GeckoEidechse: Whoops, no. Removed in 0044e58e0df0bc341e2fbbe28ce3b34985ae51b0

PR #16, opened 2022-10-15
feat: Add persistent store

GeckoEidechse:
Adds persistent store using [`tauri-plugin-store`](https://github.com/tauri-apps/tauri-plugin-store).

For now, it's only used to store the selected release canal.

### DIFF
--- a/src-vue/package.json
+++ b/src-vue/package.json
@@ -11,6 +11,7 @@
   "dependencies": {
     "@element-plus/icons-vue": "^2.0.9",
     "element-plus": "^2.2.17",
+    "tauri-plugin-store-api": "github:tauri-apps/tauri-plugin-store#dev",
     "vue": "^3.2.37",
     "vuex": "^4.0.2"
   },
--- a/README.md
+++ b/README.md
@@ -61,6 +61,37 @@ Note that you can adjust the behaviour of Tauri windows in `tauri.conf.json`, e.
 ]
 ```
 
+### Docs
+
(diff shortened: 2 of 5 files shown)


src-vue/package.json
```diff
@@ -11,6 +11,7 @@
   "dependencies": {
     "@element-plus/icons-vue": "^2.0.9",
     "element-plus": "^2.2.17",
+    "tauri-plugin-store-api": "github:tauri-apps/tauri-plugin-store#dev",
```
Alystrasz, line 14: The #dev means npm should use `dev` branch?

Alystrasz:
Also, I think we need a readme part where we explain difference between different application stores

src-vue/package.json
GeckoEidechse: Yup, there's no tagged release yet unfortunately: https://github.com/tauri-apps/tauri-plugin-store

I basically just did `npm install https://github.com/tauri-apps/tauri-plugin-store\#dev`

GeckoEidechse:
> Also, I think we need a readme part where we explain difference between different application stores

Where should we put it? Add it to the one in `src-vue`? Cause the readme in repo root is too end-user facing IMO ^^

src-vue/package.json
Alystrasz: Aren't there more stable branches to rely on?

Alystrasz:
Well, GitHub is a development platform after all :)
We could put it under the `Development` section.

PR #17, opened 2022-10-15
feat: Initial backend code to get list of installed mods

GeckoEidechse:
For now simply parses `enabledmods.json`.
Note that this file will not be up-to-date if the user just installed a mod but hasn't launched Northstar yet.

### DIFF
--- a/src-vue/src/views/ModsView.vue
+++ b/src-vue/src/views/ModsView.vue
@@ -1,18 +1,36 @@
 <template>
     <div class="fc__mods__container">
         <h3>Installed Mods:</h3>
-        TODO
+        <div>
+            <el-card shadow="hover" v-for="mod in releases">
+                <el-switch style="--el-switch-on-color: #13ce66; --el-switch-off-color: #8957e5" v-model="mod.enabled" disabled />
+                {{mod.name}}
+            </el-card>
+        </div>
     </div>
 </template>
 
 <script lang="ts">
 import { defineComponent } from "vue";
 import { ElNotification } from "element-plus";
+import { invoke } from '@tauri-apps/api/tauri';
+import { GameInstall } from "../utils/GameInstall";
 
 export default defineComponent({
     name: "ModsView",
-    methods: {
-
+    data() {
+        return {
+            releases: [],
+        }
+    },
+    async mounted() {
+        let game_install = {
+            game_path: this.$store.state.game_path,
+            install_type: this.$store.state.install_type
+        } as GameInstall;
+        // Call back-end for installed mods
+        this.releases = await invoke("get_installed_mods_caller", { gameInstall: game_install });
+        console.log(this.releases);
     }
 });
 </script>
--- a/src-vue/src/views/ModsView.vue
+++ b/src-vue/src/views/ModsView.vue
@@ -1,0 +1,25 @@
+<template>
+    <div class="fc__mods__container">
+        <h3>Installed Mods:</h3>
+        TODO
+    </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from "vue";
+import { ElNotification } from "element-plus";
+
+export default defineComponent({
+    name: "ModsView",
+    methods: {
+
+    }
+});
+</script>
+
+<style scoped>
+.fc__mods__container {
+    padding: 20px 30px;
+    color: white;
+}
+</style>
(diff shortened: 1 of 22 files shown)


GeckoEidechse:
All the CSS changes are essentially to get scrolling working in mods view without scrolling the whole window.

src-vue/src/views/ModsView.vue
```diff
@@ -0,0 +1,43 @@
...
+    name: "ModsView",
+    data() {
+        return {
+            releases: [],
+        }
+    },
+    async mounted() {
+        let game_install = {
+            game_path: this.$store.state.game_path,
+            install_type: this.$store.state.install_type
```
GeckoEidechse, line 29: For some reason, on initial app launch this is still `null`, calling the function call below to fail.

GeckoEidechse:
So this would be ready to merge (apart from merge conflict) except that for some reason https://github.com/GeckoEidechse/FlightCore/blob/2ad6aa1006efa5cb7fa2ff5de177833eea30c4ff/src-vue/src/views/ModsView.vue#L27-L46 is called already on application launch and not when opening the view.

As such it has a race condition with finding install path which it tends to win and as such then throws an error cause `this.$store.state.game_path` is null at that time.

My assumption was that code in `mounted()` would only be run when actually opening the view unless I'm mistaking something here...
Any ideas @Alystrasz? ^^

Alystrasz:
> My assumption was that code in mounted() would only be run when actually opening the view unless I'm mistaking something here...

Since our component views are tabs, they're all mounted at application start; even if they're not visible, they still are built.

This might show there's a design flaw in the app; why shouldn't we able to call Rust methods without arguments at all? Backend is supposed to have all information (`game_path`, `install_type`...)

GeckoEidechse:
> This might show there's a design flaw in the app; why shouldn't we able to call Rust methods without arguments at all? Backend is supposed to have all information already (`game_path`, `install_type`...)

Cause the information is stored in frontend not backend. The design is for frontend to handle UI and storing info and backend functions are called to execute various tasks.

Keeping information in frontend makes sense cause it means we don't need any global data store or perform synchronisation/locking of variables in the backend. All threads in backend can therefore be completely independent, meaning there's no need for locking at all.

In theory I could change `get_installed_mods()` to re-discover game path if no info was passed but that breaks when the game install cannot be auto-discovered, e.g. when installed with Lutris on Linux... ._.

Alystrasz:
> Keeping information in frontend makes sense cause it means we don't need any global data store or perform synchronisation/locking of variables in the backend. All threads in backend can therefore be completely independent, meaning there's no need for locking at all.

This however leads to inconsistencies between Rust state and UI state, which I foresee a lot of bugs will come from :smiling_face_with_tear:

GeckoEidechse:
> This however leads to inconsistencies between Rust state and UI state, which I foresee a lot of bugs will come from

Well good news, there's absolutely zero state in Rust currently :P

GeckoEidechse:
> > Well good news, there's absolutely zero state in Rust currently :P
> 
> The issue here is _currently_ :)
> 
> I think I'd prefer a stupid UI which only displays data and have logic written in Rust, compared to an UI that consistently have to call backend with same arguments

I don't think that's feasible. The UI needs to have some state e.g. for displaying the current tab, or what to show on the omni-button.
Having the state stored in frontend means we're getting this for free. The alternative would be to refactor the whole codebase to emit events on state changes and adding various forms of reference counting and locking for global variables (or using a shared object controlled by some thread). I think we can both agree on that storing state in both front and backend is bad due to needing to sync etc. Furthermore we would also have to redo persistence cause that is currently done via frontend plugin.

So given that currently frontend is already the one holding the state while backend is stateless, I think it's easier to just adjust this PR to load mods when switching to _Mods_ tab (or when game install has been detected) :P

Alystrasz:
> So given that currently frontend is already the one holding the state while backend is stateless, I think it's easier to just adjust this PR to load mods when switching to Mods tab (or when game install has been detected) :P

It's easier to just adjust this PR, of course :)

You could listen to `tab-click` events (https://element-plus.org/en-US/component/tabs.html#tabs-events), and load your content only when mods view is displayed.

GeckoEidechse:
Closing as implemented with #27 and #28

PR #21, opened 2022-10-19
refactor: Use `Result<>` return type for Linux checks

GeckoEidechse:
Allows for comprehensive error messages depending on which check failed (we only have one check right now but that will change in the future).


GeckoEidechse:
@TH3-S4LM0N cause it touches stuff you wrote ^^

(GitHub won't let me request review from you, so I'm just pinging you instead :P)

TH3-S4LM0N:
Looks fine to me 👍

PR #25, opened 2022-10-20
refactor: Router view

Alystrasz:
Rewrites application views system so it uses a router; this allows to lazy-load components (= not mounting them only corresponding menu item is clicked).


GeckoEidechse (commented):
Seems to work fine in testing for the most part. `mounted()` no only gets called when the view is actually opened, i.e. what we wanted to achieve with this PR ^^

One issue I noticed is that when enabling/disabling dev mode, the highlighted tab at the top gets updated incorrectly.
In particular, enabling dev mode highlights the _DEV_ tab even though it isn't the open one:
screenshot

and disabling dev mode switches back to _PLAY_ view but doesn't highlight it:
screenshot

PR #27, opened 2022-10-20
feat: Add initial skeleton for ModsView

GeckoEidechse:
Precedes #17

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -26,6 +26,12 @@ pub struct GameInstall {
     pub install_type: InstallType,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct NorthstarMod {
+    pub name: String,
+    pub enabled: bool,
+}
+
 /// Check version number of a mod
 pub fn check_mod_version_number(path_to_mod_folder: String) -> Result<String, anyhow::Error> {
     // println!("{}", format!("{}/mod.json", path_to_mod_folder));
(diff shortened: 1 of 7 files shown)


src-tauri/src/lib.rs
```diff
@@ -471,3 +477,37 @@ pub fn set_mod_enabled_status(
...
+    // Check if valid JSON and parse
+    let res: serde_json::Value = match serde_json::from_str(&data) {
+        Ok(res) => res,
+        Err(err) => return Err(err.to_string()),
+    };
+
+    let mut installed_mods = Vec::new();
+
+    for (key, value) in res.as_object().unwrap() {
+        println!("{:?} ===> {:?}", key, value);
```
Alystrasz, line 500: You should remove this.

src-vue/src/views/DeveloperView.vue
```diff
@@ -112,6 +116,33 @@ export default defineComponent({
...
                     });
                 });
+        },
+        async getInstalledMods() {
+            let game_install = {
+                game_path: this.$store.state.game_path,
+                install_type: this.$store.state.install_type
+            } as GameInstall;
+            await invoke("get_installed_mods_caller", { gameInstall: game_install }).then((message) => {
+                // Simply console logging for now
```
Alystrasz, line 126: So this will get removed in next PR, right?

src-tauri/src/lib.rs
```diff
@@ -471,3 +477,37 @@ pub fn set_mod_enabled_status(
 
     Ok(())
 }
+
+/// Gets list of installed mods and their properties
+/// - name
+/// - is enabled?
+pub fn get_installed_mods(game_install: GameInstall) -> Result<Vec<NorthstarMod>, String> {
+    let enabled_mods_json_path = format!("{}/R2Northstar/enabledmods.json", game_install.game_path);
```
Alystrasz, line 485: How is this working on Windows with wrong path separators?

Alystrasz (changes requested):

src-tauri/src/lib.rs
GeckoEidechse: Idk, it kinda just does ¯\\\_(ツ)\_/¯

And this is not the first time I've used forward slashes in the codebase:
https://github.com/GeckoEidechse/FlightCore/blob/5cc82fb80f5f91ed9d4a32a71e090a9a805ff2a9/src-tauri/src/lib.rs#L208


We could look into using the right path separator for the right OS but for this PR I think it's fine to keep using forward slashes. ^^

src-vue/src/views/DeveloperView.vue
GeckoEidechse: Not the next one directly but once the feature is stabilised and release, then yes.

src-tauri/src/lib.rs
GeckoEidechse: Done with 6148dd0945786b93473b5f58201b0671bcd2ec96

src-tauri/src/lib.rs
Alystrasz: If it works, it works™️

PR #28, opened 2022-10-22
feat: Show installed mods

GeckoEidechse:
Simply parses `enabledmods.json`. In the future we should also opt to check individual mods and compare with the JSON file.

This is #17 but with only the essential CSS changes to make the mod list properly scrollable.

### DIFF
--- a/src-vue/src/views/ModsView.vue
+++ b/src-vue/src/views/ModsView.vue
@@ -1,15 +1,50 @@
 <template>
     <div class="fc__mods__container">
         <h3>Installed Mods:</h3>
-        TODO
+        <div>
+            <el-card shadow="hover" v-for="mod in installed_mods">
+                <el-switch style="--el-switch-on-color: #13ce66; --el-switch-off-color: #8957e5" v-model="mod.enabled"
+                    disabled />
+                {{mod.name}}
+            </el-card>
+        </div>
     </div>
 </template>
 
 <script lang="ts">
 import { defineComponent } from "vue";
+import { ElNotification } from "element-plus";
+import { invoke } from '@tauri-apps/api/tauri';
+import { GameInstall } from "../utils/GameInstall";
+import { NorthstarMod } from "../utils/NorthstarMod"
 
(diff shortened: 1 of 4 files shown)


src-vue/src/views/ModsView.vue
```diff
@@ -1,15 +1,50 @@
...
+            install_type: this.$store.state.install_type
+        } as GameInstall;
+        // Call back-end for installed mods
+        await invoke("get_installed_mods_caller", { gameInstall: game_install })
+            .then((message) => {
+                this.installed_mods = (message as NorthstarMod[]);
+            })
+            .catch((error) => {
+                console.error(error);
+                ElNotification({
```
Alystrasz, line 40: It's not very cool to directly display errors to end-users.
What errors do you expect here?

Alystrasz (changes requested):

src-vue/src/views/ModsView.vue
GeckoEidechse: Error is basically whatever the backend returns. Showing the backend error message unprocessed makes sense cause this allows us to show a detailed error message depending on the error (i.e. missing `enabledmods.json` vs incorrectly structured`enabledmods.json`).


Examples:

`enabledmods.json`  doesn't exist:

image

Invalid `enabledmods.json`:

image


Of course we could (should?) adjust the message that the backend actually sends but I'd do that in a separate PR once I also look into removing all the `.expect`s and `.unwrap`s. ^^

Alystrasz:
Got it working on Linux, LGTM.
Last question before merging, is `enabledmods.json` file required for Northstar to operate?

GeckoEidechse:
> Last question before merging, is `enabledmods.json` file required for Northstar to operate?

Nope. It generates it, if it doesn't exist.

Using `enabledmods.json` to get the list of installed mods is also not a good idea as deleted mods are never removed from that last list but this is just a first implementation, I'm changing the source for the list of installed mods in another PR in the near future ^^

PR #32, opened 2022-11-01
feat: Initial support for installing mods from TS

GeckoEidechse:
This is the basic code needed to install a mod from Thunderstore.

There's an issue where the removal of the temporary download folder does not work properly which I have yet to resolve.

- [x] Basic install logic
- [x] Prevent installing `Northstar` or `NorthstarReleaseCandidate`
- [x] Handle dependencies -> Waiting on `libthermite` update as current is broken.
- [x] Add Thunderstore dependency string to `mod.json` (for checking for updates).


src-tauri/src/lib.rs
```diff
@@ -510,3 +510,75 @@ pub fn get_installed_mods(game_install: GameInstall) -> Result<Vec<NorthstarMod>
 
     Ok(installed_mods)
 }
+
+async fn get_ns_mod_download_url(thunderstore_mod_string: String) -> Result<String, String> {
+
+    // TODO: This will crash the thread if not internet connection exist. `match` should be used instead
```
Alystrasz, line 516: So this is not ready for review yet 😄

src-tauri/src/lib.rs
```diff
@@ -510,3 +510,75 @@ pub fn get_installed_mods(game_install: GameInstall) -> Result<Vec<NorthstarMod>
...
+        Ok(pkg) => pkg,
+        Err(err) => return Err(err.to_string()),
+    };
+    dbg!(pkg.clone());
+
+    // Delete downloaded zip file
+    std::fs::remove_file(path).unwrap();
+
+    // Delete temp download folder
+    // TODO: For some reason this fails. Maybe delete download folder in separate call?
```
Alystrasz, line 579: Why doesn't this work?

src-vue/src/views/DeveloperView.vue
```diff
@@ -143,6 +157,34 @@ export default defineComponent({
...
+        async installMod() {
+            let game_install = {
+                game_path: this.$store.state.game_path,
+                install_type: this.$store.state.install_type
+            } as GameInstall;
+            let mod_to_install = this.mod_to_install_field_string;
+            await invoke("install_mod_caller", { gameInstall: game_install, thunderstoreModString: mod_to_install }).then((message) => {
+                // Simply console logging for now
+                // In the future we should display the installed mods somewhere
+                console.log(message);
```
Alystrasz, line 170: This should be removed

Alystrasz (changes requested):

src-vue/src/views/DeveloperView.vue
```diff
@@ -143,6 +157,34 @@ export default defineComponent({
...
+        },
+        async installMod() {
+            let game_install = {
+                game_path: this.$store.state.game_path,
+                install_type: this.$store.state.install_type
+            } as GameInstall;
+            let mod_to_install = this.mod_to_install_field_string;
+            await invoke("install_mod_caller", { gameInstall: game_install, thunderstoreModString: mod_to_install }).then((message) => {
+                // Simply console logging for now
+                // In the future we should display the installed mods somewhere
```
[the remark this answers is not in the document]
GeckoEidechse: Yup cause the mod list in FlightCore right now only uses `enabledmods.json`. Switching up the method for grabbing mods to reading out the `mod.json` for all folders in the `mods` directory should fix this but that's unrelated to this PR ^^

src-tauri/src/main.rs
```diff
@@ -303,3 +305,12 @@ fn disable_all_but_core_caller(game_install: GameInstall) -> Result<(), String>
 async fn get_installed_mods_caller(game_install: GameInstall) -> Result<Vec<NorthstarMod>, String> {
     get_installed_mods(game_install)
 }
+
+#[tauri::command]
+/// Installs the specified mod
+async fn install_mod_caller(
```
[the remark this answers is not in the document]
GeckoEidechse: Not sure if that's actually necessary... For one it needs to compare versions cause we definitely wanna allow install for newer versions of the same mod as well as maybe a force-reinstall in cases where the user may have changed the mod files and wants to force-reinstall.

I feel like it might make more sense to handle this in the frontend in the sense that it will have a button that shows `Install`, `Force reinstall`, or `Update` depending on (the existence of) the locally installed mod.

src-tauri/src/lib.rs
```diff
@@ -510,3 +510,75 @@ pub fn get_installed_mods(game_install: GameInstall) -> Result<Vec<NorthstarMod>
...
+        }
+    }
+
+    Err("Could not find mod on Thunderstore".to_string())
+}
+
+// Copied from `libtermite` source code and modified
+// Should be replaced with a library call to libthermite in the future
+/// Download and install mod to the specified target.
+pub async fn fc_install_mod(
```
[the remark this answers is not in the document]
GeckoEidechse: Uh, do you know more than me? I intended for it to just download and install the given mod :P

src-tauri/src/lib.rs
GeckoEidechse: Eh, FlightCore will already do that if you launch it without internet :>

src-tauri/src/lib.rs
GeckoEidechse: Uncommenting it and then installing a mod gives:

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 303, kind: Uncategorized, message: "File cannot be opened cause it is currently being deleted." }', E:\Dateien\Coding\FlightCore\src-tauri\src\lib.rs:581:49
```

(translated error message from German by hand)

src-tauri/src/lib.rs
Alystrasz: I'll open a dedicated issue for this

src-tauri/src/lib.rs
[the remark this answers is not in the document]
Alystrasz: It also fetches an archive from Thunderstore :)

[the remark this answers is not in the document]
GeckoEidechse: Done with 66007b91eb2fbe5805821ba23644a98586385f02

src-vue/src/views/DeveloperView.vue
GeckoEidechse: Done with 66007b91eb2fbe5805821ba23644a98586385f02

src-vue/src/views/DeveloperView.vue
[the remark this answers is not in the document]
GeckoEidechse: Will be resolved with #38 in the future.

src-tauri/src/main.rs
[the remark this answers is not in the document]
Alystrasz: Fair!

src-tauri/src/lib.rs
Alystrasz: Can't you just remove the `___flightcore-temp-download-dir/` directory?

GeckoEidechse: What do you mean exactly?

Delete it after done installing mod? Cause that's what the commented out function is supposed to do.

Don't use the temp folder in the first place? Not a big fan cause a crash mid-install will cause leftover files littering the Titanfall2 directory as opposed to being restricted to a single folder ^^

GeckoEidechse:
Reminder to self that it doesn't do dependency install yet (and it kinda should before making it proper public).

src-tauri/src/lib.rs
Alystrasz: Yeah, I mean only deleting directory after mod installation; I suggest to keep only the `std::fs::remove_dir_all(download_directory).unwrap();` call.

While commenting all remove calls, I sometimes have another exception:
```rust
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: StripPrefixError(())', C:\Users\Remy\.cargo\registry\src\github.com-1ecc6299db9ec823\libthermite-0.2.0\src\core\actions.rs:197:46
```

`The_Peepeepoopoo_man-Chimera-0.1.1` causes this issue (creating a `1667515294` folder with zip content in my `mods` directory), while `sammydoodlew-Volt_Vaporwave-1.1.0` seems to install just fine.

GeckoEidechse: > Yeah, I mean only deleting directory after mod installation; I suggest to keep only the `std::fs::remove_dir_all(download_directory).unwrap();` call.

Both directory remove calls don't work individually, I had them both in there still as comments as a reminder to me. Didn't intend to use both :P

> ```rust
> thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: StripPrefixError(())', C:\Users\Remy\.cargo\registry\src\github.com-1ecc6299db9ec823\libthermite-0.2.0\src\core\actions.rs:197:46
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

In regards to that panic is in `libthermite`, might need to check with Emerald on the issue with the lib (although our version of the lib is outdated atm).

Another thing I realised, is that if we do multiple mod installs at once e.g. user clicking on update buttons for multiple mods, deleting the folder in the install call will delete the folder while the second mod install is still going.

So maybe we should just resort to deleting temp download folder in a separate function call. ^^

Alystrasz: > So maybe we should just resort to deleting temp download folder in a separate function call. ^^

Nah, this method should remove all folders it had created.
A simple solution is not to use same name - for all temporary folders, but rather to use mod name and/or current timestamp.

GeckoEidechse: That would work. My main concern is leftover files from failed operations...

So maybe a common prefix (the same one we use rn) that could be used to identify folder/files created by FlightCore, so they can be easily removed again if necessary?

AnActualEmerald: > In regards to that panic is in libthermite, might need to check with Emerald on the issue with the lib (although our version of the lib is outdated atm).

Published v0.3.2 this morning, that panic now returns an error and the temp directory `install_mod` creates should now only exist in the scope of the function

Alystrasz: @GeckoEidechse will updating `libthermite` solve this issue?

GeckoEidechse: It should yes. But before updating `libthermite` I'd like to get #38 merged and then make a new release so that I can better track down any errors, should they occur.

GeckoEidechse: With `0.3.3` of `libthermite` installing `The_Peepeepoopoo_man-Chimera-0.1.1` now throws a proper error message ^^

image

GeckoEidechse: Given that zip file gets removed and all that is left-over is an empty folder, I have decided that this is a non-issue, to be dealt with by future-Gecko™ in a separate PR ¯\\\_(ツ)\_/¯

Alystrasz: That definitely is an issue, though. :)

GeckoEidechse: Done with c74de4fa7bbceb019931bae3244acd55e3114b8c

GeckoEidechse:
@Alystrasz (properly) addressed all mentioned issues now. Anything else before merge? ^^

Alystrasz:
Yep, I want to be able to run this code before merging it in, which will have to wait a bit since I'm on a business trip in Switzerland.

Alystrasz:
I managed to install Titanfall2 on my Ubuntu (discovering Lutris by the same occasion :)).

Installing `The_Peepeepoopoo_man-Chimera-0.1.1`, I get a success message, but the mod doesn't appear in the mods folder nor the mods view.
image
with the log
```text
[src/mod_management/mod.rs:187] thunderstore_mod_string.clone() = "The_Peepeepoopoo_man-Chimera-0.1.1"
[src/mod_management/mod.rs:200] ns_mod.clone() = ModVersion {
    name: "Chimera",
    version: "0.1.1",
    url: "https://thunderstore.io/package/download/The_Peepeepoopoo_man/Chimera/0.1.1/",
    desc: "Custom titan demonstrating titanFramework 0.3s custom passives",
    deps: [
        "northstar-Northstar-1.9.4",
        "The_Peepeepoopoo_man-Titanframework-0.3.0",
    ],
    installed: false,
    global: false,
    file_size: 71515,
}
[src/mod_management/mod.rs:232] deps.clone() = [
    "northstar-Northstar-1.9.4",
    "The_Peepeepoopoo_man-Titanframework-0.3.0",
]
[src/mod_management/mod.rs:187] thunderstore_mod_string.clone() = "northstar-Northstar-1.9.4"
[src/mod_management/mod.rs:200] ns_mod.clone() = ModVersion {
    name: "Northstar",
    version: "1.9.4",
    url: "https://thunderstore.io/package/download/northstar/Northstar/1.9.4/",
    desc: "Titanfall 2 modding and custom server framework.",
    deps: [],
    installed: false,
    global: false,
    file_size: 73374598,
}
[src/mod_management/mod.rs:232] deps.clone() = []
Failed to delete download folder due to Aucun fichier ou dossier de ce type (os error 2)
```

---

On the other hand, `sammydoodlew-Volt_Vaporwave-1.1.0` seems to install just fine!

GeckoEidechse:
> Installing `The_Peepeepoopoo_man-Chimera-0.1.1`, I get a success message, but the mod doesn't appear in the mods folder nor the mods view.

Huh, good catch. Can reproduce on my side as well so I'll investigate.

Side note: For testing on laptop I just created an empty `Titanfall.exe` file and point FlightCore to it xD

GeckoEidechse:
> Installing `The_Peepeepoopoo_man-Chimera-0.1.1`, I get a success message, but the mod doesn't appear in the mods folder nor the mods view.

Fixed with fb625adeaffe37096bb2ab6d1a0a54382d3158c3 in a single line :,)

src-tauri/src/mod_management/mod.rs
```diff
@@ -130,3 +131,183 @@ pub fn get_installed_mods_and_properties(
...
+                println!("Failed setting modstring for {}", nsmod.name);
+                println!("{}", err);
+            }
+        }
+    }
+
+    // Delete downloaded zip file
+    std::fs::remove_file(path).unwrap();
+
+    // Delete temp download folder
```
Alystrasz, line 307: I suppose you can remove this before merge :)

Alystrasz (changes requested):

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Oh yeah, removed with 29d1c644c39283f39a524b0517f3868150b58d1c o7

AnActualEmerald (approved):
LGTM

Alystrasz (approved):

Issue #33, opened 2022-11-01
No internet?


Alystrasz:
Lots of methods currently fail if there's no internet connection.

Alystrasz:
App currently prints out HTTP errors instead of crashing.

PR #35, opened 2022-11-02
refactor: Make all functions async

GeckoEidechse:
This way none of them are run on main thread and as such if a function panics, it will only crash that specific thread allowing the rest of the application to survive.

The only exception is `force_panic()` as its purpose is hard crashing the application.

From the frontend side of things, `invoke()` is already async, so no frontend changes are needed.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -97,7 +97,7 @@ fn main() {
 
 #[tauri::command]
 /// Wrapper for `find_game_install_location` as tauri doesn't allow passing `Result<>` types to front-end
-fn find_game_install_location_caller() -> Result<GameInstall, String> {
+async fn find_game_install_location_caller() -> Result<GameInstall, String> {
     match find_game_install_location() {
         Ok(game_install) => Ok(game_install),
         Err(err) => {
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Sorta resolves #33 in the sense that now the application no longer crashes. We should of course still properly handle no-internet scenarios instead of just letting that thread crash. That's outside the scope of this PR though :P

GeckoEidechse:
As a note to self, `async` functions that panic are still caught by sentry.io integration and log properly.

Alystrasz (approved):

Alystrasz:
Oops @GeckoEidechse, didn't notice it, but this seems to introduce an error with some asynchronous code.

```shell
  VITE v3.1.3  ready in 443 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
        Info Watching /home/remy/Travail/main/FlightCore/src-tauri for changes...
    Blocking waiting for file lock on package cache
   Compiling app v0.2.3 (/home/remy/Travail/main/FlightCore/src-tauri)
    Finished dev [unoptimized + debuginfo] target(s) in 58.79s
Checking GitHub API
thread 'tokio-runtime-worker' panicked at 'Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.', /home/remy/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.1/src/runtime/blocking/shutdown.rs:51:21
Couldn't locate Steam on this computer!
Could not auto-detect game install location! Please enter it manually.
```

PR #38, opened 2022-11-02
feat: Parse actually installed mods

GeckoEidechse:
WIP


Alystrasz:
@GeckoEidechse fuck I also did the merging part
Could you move the new methods inside the new `mod_management` package?

GeckoEidechse:
> Could you move the new methods inside the new `mod_management` package?

Done o7

Alystrasz:
So this is supposed to load mods list even if there's no `enabledmods.json`, right?
If yes, it's not working on my end.

GeckoEidechse:
> So this is supposed to load mods list even if there's no `enabledmods.json`, right? If yes, it's not working on my end.

Good catch. It should work even with `enabledmods.json` missing. Fixed in f3d2e63e6a8922bce9f40b6c00c11df5ff4c0850

GeckoEidechse:
@Alystrasz anything else or is the PR ready to merge? ^^

Alystrasz:
> @Alystrasz anything else or is the PR ready to merge from your side? ^^

I still have to try it on my machine, I'll have time this evening (I hope) to review again!

Alystrasz:
When user doesn't have `enabledmods.json` on his computer, FC correctly shows up the mods, but it's not possible to disable them (all appear enabled).

src-tauri/src/mod_management/mod.rs
```diff
@@ -31,3 +35,96 @@ pub fn set_mod_enabled_status(
...
+
+/// Parses `mod.json` for mod name
+// TODO: Maybe pass PathBuf or serde json object
+fn parse_mod_json_for_mod_name(mod_json_path: String) -> Result<String, anyhow::Error> {
+    // Read file into string and parse it
+    let data = std::fs::read_to_string(mod_json_path)?;
+    let parsed_json: serde_json::Value = serde_json::from_str(&data)?;
+
+    // Extract mod name
+    let mod_name = match parsed_json.get("Name").and_then(|value| value.as_str()) {
```
Alystrasz, line 48: We should enforce somewhere (Northstar-side) what is a correct mod format; I fear the `"name"` instead of `"Name"` issue here.

src-tauri/src/mod_management/mod.rs
```diff
@@ -31,3 +35,96 @@ pub fn set_mod_enabled_status(
 
     Ok(())
 }
+
+
+/// Parses `mod.json` for mod name
+// TODO: Maybe pass PathBuf or serde json object
```
Alystrasz, line 41: Still TODO?

src-tauri/src/mod_management/mod.rs
```diff
@@ -31,3 +35,96 @@ pub fn set_mod_enabled_status(
...
+pub fn get_installed_mods_and_properties(
+    game_install: GameInstall,
+) -> Result<Vec<NorthstarMod>, String> {
+    // Get actually installed mods
+    let found_installed_mods = get_installed_mods(game_install.clone())?;
+
+    // Get enabled mods as JSON
+    let enabled_mods: serde_json::Value = match get_enabled_mods(game_install) {
+        Ok(enabled_mods) => enabled_mods,
+        Err(_) => serde_json::from_str("{}").unwrap(), // `enabledmods.json` not found, create empty object
```
Alystrasz, line 110: We should probably create an object with all mods as entries, to solve https://github.com/GeckoEidechse/FlightCore/pull/38#issuecomment-1309080761

Alystrasz (changes requested):

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Yup but not for this PR IMO. This would basically just give negligible performance increase and slightly better looking code but I don't have enough experience with `PathBuf` object yet to tackle this in this PR.

GeckoEidechse:
> When user doesn't have `enabledmods.json` on his computer, FC correctly shows up the mods, but it's not possible to disable them (all appear enabled).

This is sorta "intended behaviour" in the sense that the issue is that we cannot yet set the value of an entry in `enabledmods.json` that doesn't exists yet. I'm planning to address that in a next PR but given that rn an end-user cannot install a mod using FlightCore, they'll barely run into this issue before I'll address it ^^

As for the being enabled part if entry doesn't exists in `enabledmods.json` (or the file doesn't exist at all), that basically mirrors the behaviour in Northstar. Maybe `enabledmods.json` should be called `disabledmods.json` by Northstar :P

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Eh the object feels kinda overkill, could do but not really necessary and won't solve https://github.com/GeckoEidechse/FlightCore/pull/38#issuecomment-1309080761. That issue is addressed in https://github.com/GeckoEidechse/FlightCore/pull/38#issuecomment-1309493671

src-tauri/src/mod_management/mod.rs
GeckoEidechse: This seems to be enforced to `"Name"` based on NorthstarLauncher source code: https://github.com/R2Northstar/NorthstarLauncher/blob/6f46d0ec44e95252b2e4fa08bd48290383ea1d59/NorthstarDLL/modmanager.cpp#L58

Sidenote: I'm planning to do a mods sanity checker but that's kinda outside of the scope of the PR.

Alystrasz (approved):

GeckoEidechse:
Dope, time to make next release :D

PR #42, opened 2022-11-07
style: Add `.editorconfig` file

GeckoEidechse:
Describes some basic formatting like indentation with spaces, indentation size, final newlines, etc.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# For more info on `.editorconfig` file see https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*.{rs,ts,vue,css}]
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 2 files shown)


Alystrasz:
* How is this applied to source code?
* Is is applied to Typescript or Rust or both?

GeckoEidechse:
> * How is this applied to source code?

It's technically applied to the IDE, so that when it uses certain autoformatting options, it takes the info from `.editorconfig` into account. The config file itself is supported by [a variety of IDEs either with or without extension](https://editorconfig.org/#pre-installed)-

> * Is is applied to Typescript or Rust or both?

[Both](https://github.com/GeckoEidechse/FlightCore/pull/42/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5bR6), though specific rules can be added for each.

Alystrasz:
It seems to work with Visual Studio Code.
You should add a note to developers mentioning they should install an EditorConfig plugin in their IDE, otherwise LGTM :)

GeckoEidechse:
> It seems to work with Visual Studio Code. You should add a note to developers mentioning they should install an EditorConfig plugin in their IDE, otherwise LGTM :)

Added note with 8e419dbb8bcc8c439eb8afd791c56ba7d9cdbd76

PR #44, opened 2022-11-10
refactor: Clicking FC version activates dev

GeckoEidechse:
Clicking FC version activates dev instead of clicking Northstar version.

Makes dev mode a bit more "hidden" (i.e. less likely to be accidentally opened by end-user) and it's more logical to click FlightCore version to open **FlightCore** dev mode as opposed to clicking Northstar version.


Alystrasz (approved):
I don't see how people would click Northstar version indicator 6 times in a row by accident, but you're right, we never know :)

PR #45, opened 2022-11-10
feat: Add categories for dev mode buttons

GeckoEidechse:
Makes buttons a bit more organised ^^

screenshot


Alystrasz (approved):
What a nice layout!

PR #47, opened 2022-11-12
refactor: Splice out GH release API fetch function

GeckoEidechse:
So that it can be re-used by other functions later such as for checking newest FlightCore version by checking against GitHub releases.


Alystrasz (approved):
LGTM

PR #48, opened 2022-11-12
feat: Show stripped title bar on debug build

GeckoEidechse:
Helps with differentiating which window is which when both developing FlightCore and is using release version at the same time.
Wouldn't be the first time I accidentally nearly confused the two :P

Rendered result:
image

### DIFF
--- a/src-vue/src/plugins/store.ts
+++ b/src-vue/src/plugins/store.ts
@@ -200,6 +200,12 @@ async function _initializeApp(state: any) {
     // Enable dev mode directly if application is in debug mode
     if (await invoke("is_debug_mode")) {
         state.developer_mode = true;
+
+        // Make menubar striped if debug build
+        let menu_bar_handle = document.querySelector('#fc__menu_bar');
+        if (menu_bar_handle !== null) {
+            menu_bar_handle.classList.toggle('developer_build');
+        }
     }
 
     // Grab Northstar release canal value from store if exists
--- a/src-vue/src/App.vue
+++ b/src-vue/src/App.vue
@@ -41,7 +41,6 @@ export default {
         default-active="/"
         router
         mode="horizontal"
-        class="developer_build"
         id="fc__menu_bar"
         data-tauri-drag-region
     >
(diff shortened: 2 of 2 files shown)


GeckoEidechse:
I just have no idea how to apply that CSS class only if [`await invoke("is_debug_mode")` returns true](https://github.com/GeckoEidechse/FlightCore/blob/f1756396ecba4ace6d62ee94e03d060c8f337de3/src-vue/src/plugins/store.ts#L201)...

@Alystrasz any ideas? ^^

Alystrasz:
Love the CSS style ❤️ 
When App is mounted, you can add the CSS class dynamically, regarding invoke result:
```typescript
document.querySelector('#fc__menu_bar').classList.toggle('developer_build');
```
An even cleaner way to do this is to apply a [`ref` attribute](https://vuejs.org/guide/essentials/template-refs.html) to our menu, and referencing it as such:
```typescript
this.$refs.menu.classList.toggle('developer_build');
```

GeckoEidechse:
I opted for

```
document.querySelector('#fc__menu_bar').classList.toggle('developer_build');
```

cause I don't know how to do the other one ^^"

src-vue/src/plugins/store.ts
```diff
@@ -200,6 +200,12 @@ async function _initializeApp(state: any) {
     // Enable dev mode directly if application is in debug mode
     if (await invoke("is_debug_mode")) {
         state.developer_mode = true;
+
```
Alystrasz, line 203: Actually, this is not the good place for this code; you should rather put it where dev mod is enabled/disabled.

Alystrasz (changes requested):

src-vue/src/plugins/store.ts
GeckoEidechse: > Currently, when disabling dev mod, stripped style stays there.

Uh, that's the point :P

The striped menubar is not tied to "developer mode" (which shows/hides DevView) but whether the binary has been compiled in debug mode (`npm run tauri dev`) or release mode (`npm run tauri build`).

So the idea is that if you have two FlightCore windows open. One is the locally compiled build you're working on rn and the other one is an installed release build, they are both visually distinct from another.

Sorry for the confusion haha ^^

Alystrasz (approved):
Working as expected.

PR #50, opened 2022-11-13
feat: Add JSON5 support, allows comments in JSON

GeckoEidechse:
Some mods have comments in their JSON. Northstar handles it fine, so we should too.

More info on JSON5: https://json5.org/


Alystrasz (approved):
LGTM

PR #51, opened 2022-11-13
refactor: Move GitHub API fetch to submodule and re-use existing function

GeckoEidechse:
One more step towards reducing code re-use ^^


Alystrasz (approved):
Tested on Linux, LGTM.

PR #54, opened 2022-11-14
feat: Thunderstore mods listing

Alystrasz:
This introduces interface components for Thunderstore mods listing.

Capture d’écran de 2022-11-14 23-14-16

* Displays information about mod
* Thunderstore mod page can be opened in default browser via information button
* Mods can be searched through (filtering by mod name / author / mod description)

##### TODOs

- [x] Responsive layout
- [x] Pagination (displaying all mods is slow on some devices) (THAT'S FOR ANOTHER TIME)
- [x] Search debounce (not to trigger search at each character input)
- [x] "Install/Installed" button text
- [x] Actual mod installation
- [x] Blur background on this view
- [x] Update view once a mod has been installed
- [x] Display "updating" instead of "installing"


GeckoEidechse:
Regarding:

> * Search throttle (not to trigger search at each character input)

I'd argue to keep it as is, as it feels really responsive to immediately update.

> * Pagination (displaying all mods is slow on some devices)

Tbh devices where it's slow also probably won't be able to run Titanfall2 anyway, so I don't think pagination is necessary. In fact it will mess with the search and require updating it.

src-vue/src/utils/thunderstore/ThunderstoreModVersion.d.ts
```diff
@@ -0,0 +1,8 @@
+export interface ThunderstoreModVersion {
+    description: string;
+    icon: string;
+    version_number: string;
+    download_url: string;
+    downloads: number;
+    date_created: string;
+}
```
GeckoEidechse, line 8:
```suggestion
}

```

Trailing new line :)

src-vue/src/views/ThunderstoreModsView.vue
```diff
@@ -0,0 +1,214 @@
...
+import {defineComponent} from 'vue';
+import {ThunderstoreMod} from "../utils/thunderstore/ThunderstoreMod";
+import { shell } from '@tauri-apps/api';
+import { ThunderstoreModVersion } from '../utils/thunderstore/ThunderstoreModVersion';
+
+export default defineComponent({
+    name: "ThunderstoreModsView",
+    /**
+     * TODO Mods should be stored in front store
+     * TODO Fetch mods with backend?
```
GeckoEidechse: For completion's sake:

As mentioned in DMs, I'm totally fine with doing it in front-end for now and moving it to backend later. As such the TODO comments should stay preserved here. 

So basically this code comment can be resolved as is and acts as a free-dopamine-button :P

Alystrasz:
> I'd argue to keep it as is, as it feels really responsive to immediately update. ^^

I can see slowdown on my laptop (AMD Ryzen 7 3750H / 16GB of RAM), so I imagine the issue is not that uncommon :)

> I don't think pagination is necessary. In fact it will mess with the search and require updating it.

There's a bit of logic to think about, yeah, but it's definitely doable.

src-vue/src/views/ThunderstoreModsView.vue
GeckoEidechse: re: 0f6e78246ca41b125785a24664d49ae0121b1909

lol, my comment came in too late xD

Alystrasz: I removed this while you were typing your comment :)

GeckoEidechse:
> - Actual mod installation

Assuming #32 gets merged first, all you'd need to do is invoke `install_mod_caller()` with Thunderstore mod string and `GameInstall` object: https://github.com/GeckoEidechse/FlightCore/blob/bbb427651da5fb2d820cee51dc8e971326bce121/src-vue/src/views/DeveloperView.vue#L168-L173

GeckoEidechse:
> > I'd argue to keep it as is, as it feels really responsive to immediately update. ^^
> 
> I can see slowdown on my laptop (AMD Ryzen 7 3750H / 16GB of RAM / Ubuntu 20.04.3 LTS), so I imagine the issue is not that uncommon :)

Might be a WebKit issue (AMD Ryzen7 1800X) on Windows / WebView2 seems to be fine ^^

At least perf doesn't differ between ThunderstoreView and other views.

(Also, update your Ubuntu install Gnome 40+ > Gnome3 :P)

GeckoEidechse:
> > I don't think pagination is necessary. In fact it will mess with the search and require updating it.
>
> There's a bit of logic to think about, yeah, but it's definitely doable.

I'd save for future PR ;)

GeckoEidechse:
> - Responsive layout

Number of columns changes with window size, so it's already working? Or is there something else you have in mind?

Alystrasz:
> > Actual mod installation

> Assuming https://github.com/GeckoEidechse/FlightCore/pull/32 gets merged first, all you'd need to do is invoke install_mod_caller() with Thunderstore mod string and GameInstall object

Cool :) 
Is there a way to check if current mod is already installed? I would like for buttons to display "Installed" instead of "Install" if that's the case.

> Number of columns changes with window size, so it's already working? Or is there something else you have in mind?

I want mod cards to be centered on the screen, *i.e.* margin space is the same to the left and to the right of the list.

GeckoEidechse:
> Is there a way to check if current mod is already installed? I would like for buttons to display "Installed" instead of "Install" if that's the case.

Not right now. I can look into it but it'll be a bit. I definitely want that feature but given that neither VTOL nor Viper have it either, I think we're fine with skipping it for this PR ;)

Alystrasz:
> Not right now. I can look into it but it'll be a bit. I definitely want that feature but given that neither VTOL nor Viper have it either, I think we're fine with skipping it for this PR ;)

Meh. I don't feel good shipping unfinished features :smiling_face_with_tear:

GeckoEidechse:
> > > I'd argue to keep it as is, as it feels really responsive to immediately update. ^^
> > 
> > 
> > I can see slowdown on my laptop (AMD Ryzen 7 3750H / 16GB of RAM / Ubuntu 20.04.3 LTS), so I imagine the issue is not that uncommon :)
> 
> Might be a WebKit issue (AMD Ryzen7 1800X) on Windows / WebView2 seems to be fine ^^ At least perf doesn't differ between ThunderstoreView and other views.
> 
> (Also, update your Ubuntu install Gnome 40+ > Gnome3 :P)

Testing on my Linux laptop (`Intel i7-7Y75` with `Intel HD Graphics 615`) the performance stays the same regardless of the number if mods displayed. So I think what we're facing here is an inherent WebKitGTK issue and less of "too many elements" kinda situation, so I have doubts that pagination will solve it.

GeckoEidechse:
From playing around a bit more with it, I can tell that on low-powered devices (unsurprisingly) native scrolling performs far better than `<el-scrollbar>`. Maybe outside of the scope of this PR but if we want to improve performance, we should do as much stuff as possible using browser/webview native things.

For scrollbars, this would mean properly locking the header in CSS and then making the rest of the content scrollable.

src-vue/src/utils/thunderstore/ThunderstoreModVersion.d.ts
GeckoEidechse, line 8: This should also export `full_name: string` as that field contains the mod string needed to be passed to the backend for mod install. ^^

image

GeckoEidechse:
> > Not right now. I can look into it but it'll be a bit. I definitely want that feature but given that neither VTOL nor Viper have it either, I think we're fine with skipping it for this PR ;)
> 
> Meh. I don't feel good shipping unfinished features :smiling_face_with_tear:

I'll add the Thunderstore mod string to the returned modstructs with #56. From there you should be able to check against it to see whether a mod is installed or not and because the mod string also contains version, you should also be able to see if it's up-to-date. ^^

src-vue/src/utils/thunderstore/ThunderstoreModVersion.d.ts
Alystrasz: Added in b8ef66662cef9d5bc58b26a07af418024c3f8f9e

GeckoEidechse:
> > * Search throttle (not to trigger search at each character input)
> 
> I'd argue to keep it as is, as it feels really responsive to immediately update. ^^

Tested 34ae8b70c87d70539cefdb68b20d9afc5d9456ef rn and I have to say I definitely prefer the old instant search on input :P

It's just very jarring to have to wait 1-2 seconds for something to load. Also current search now hides all mod cards when entering first characters when there was no previous input.

Alystrasz:
> It's just very jarring to have to wait 1-2 seconds for something to load.

My bad, I intended to set it to `200ms`, this is fixed with 1fcc47f0f6d1416c0e059336aa53f88beddd15ea.

> Also current search now hides all mod cards when entering first characters when there was no previous input.

How is this a problem?

Alystrasz:
Except this searching feature disagreement, I only see two remaining issues:
* pagination, which will arrive in a separate PR in the future
* detecting that a installed mod is outdated, I think we should do this here

GeckoEidechse:
> > It's just very jarring to have to wait 1-2 seconds for something to load.
> 
> My bad, I intended to set it to `200ms`, this is fixed with [1fcc47f](https://github.com/GeckoEidechse/FlightCore/commit/1fcc47f0f6d1416c0e059336aa53f88beddd15ea).

Okay this is surprisingly a lot better. I'd still prefer the instant search personally but we can revisit later. This is good enough for me atm ^^

> > Also current search now hides all mod cards when entering first characters when there was no previous input.
> 
> How is this a problem?

It's the abrupt change that goes from _seeing all mods_ -> _seeing no mods_ -> _seeing searched mods_ that's kinda annoying IMO. Another case where this would be an issue is when e.g. seeing a mod by a certain author, let's say `taskinoz`. You start typing in the author name into the search as it's spelled out on the card by looking at it but the moment you do, all mod cards are gone until they reappear a second later.

I'd say having it go  _seeing all mods_ -> _seeing searched mods_ instead would be preferred (at least by me :P)


> Except this searching feature disagreement, I only see two remaining issues:
> 
>     * pagination, which will arrive in a separate PR in the future

Sounds good to me, so agreed

>     * detecting that a installed mod is outdated, I think we should do this here

and agreed ^^

Alystrasz:
> It's the abrupt change that goes from seeing all mods -> seeing no mods -> seeing searched mods that's kinda annoying IMO. Another case where this would be an issue is when e.g. seeing a mod by a certain author, let's say taskinoz. You start typing in the author name into the search as it's spelled out on the card by looking at it but the moment you do, all mod cards are gone until they reappear a second later.

Okay, I'll remove the "searching mods" message when some mods are actually displayed.

> > detecting that a installed mod is outdated, I think we should do this here
> and agreed ^^

Will do then :)

GeckoEidechse:
> Okay, I'll remove the "searching mods" message when some mods are actually displayed.

That's not really the issue though. It's that on the initial search for a short period, no mod cards are displayed at all until the search completes. After that it's fine ^^"

Alystrasz:
Please make a video showing me the issue then :smile:

GeckoEidechse:
> Please make a video showing me the issue then smile

Aight got OBS working on my laptop after some fiddling ^^

https://user-images.githubusercontent.com/40122905/202862540-ade6c50f-5da1-433a-9079-19547e1ba59f.mp4

Typing in `north` when the search bar was previously empty, hides all cards until I stop typing and the search completes. Typing in `star` afterwards, the previously found cards stay present until the search completes and then the list of cards is updated.

Alystrasz:
> Okay, I'll remove the "searching mods" message when some mods are actually displayed.

Let me complete my sentence:

Okay, I'll remove the "searching mods" message when some mods are actually displayed, **and display current list of mods instead.**

Alystrasz:
Lemme quote myself from 9 days ago:

> Meh. I don't feel good shipping unfinished features :smiling_face_with_tear:

GeckoEidechse:
Ok for 

> - Display "updating" instead of "installing"

I'm fine to agree. But for background blur, if we introduce it on Thunderstore view, we'd have to do it on every view, where we have content in front of background to stay consistent. In other words, everywhere except PlayView :P

Hence my suggestion to have this point a separate PR cause it just blows up this one out of proportions ^^

Alystrasz:
Ok, so I still have ideas of new features for this:
* putting outdated mods first in the list
* allowing mods sorting by downloads count, stars count

But since you want to release v1.0.0 (and I don't want you to send me a pipe bomb), I propose to work on those later. :smile:

GeckoEidechse:
> * putting outdated mods first in the list

I assume this in regard to making it easier to update mods? Cause we could just add the update check and update button to the ModsView as well ^^

In fact we might wanna redesign the ModsView soon anyway cause compared to ThunderstoreView it now looks kinda barren :P

GeckoEidechse:
With the blurred background, there's now a brighter edge around the edges of the application when in ThunderstoreView, is that intended? ^^

image

(Testing on Windows rn)

src-vue/src/views/ThunderstoreModsView.vue
```diff
@@ -0,0 +1,190 @@
...
+                <!-- Mod cards -->
+                <thunderstore-mod-card v-for="mod of modsList" v-bind:key="mod.name" :mod="mod" />
+            </div>
+        </el-scrollbar>
+    </div>
+</template>
+
+<script lang="ts">
+import {defineComponent} from 'vue';
+import {ThunderstoreMod} from "../utils/thunderstore/ThunderstoreMod";
```
GeckoEidechse, line 32:
```suggestion
import { defineComponent } from 'vue';
import { ThunderstoreMod } from "../utils/thunderstore/ThunderstoreMod";
```

Formatting :>

src-vue/src/views/ThunderstoreModsView.vue
```diff
@@ -0,0 +1,190 @@
...
+import ThunderstoreModCard from "../components/ThunderstoreModCard.vue";
```
GeckoEidechse, line 33: Shouldn't this be:

```suggestion
import { ThunderstoreModCard } from "../components/ThunderstoreModCard.vue";
```
or am I missing something? ^^"

GeckoEidechse (commented):
Aight, looks like it's mostly formatting related things and some clarifications about stuff I'm unsure about.

There's also the blurred white border issue (https://github.com/GeckoEidechse/FlightCore/pull/54#issuecomment-1326871411), not sure how to fix that ^^

Given my lack of experience in frontend dev, my view tends to get fuzzy the moment I look at CSS so I'm just trusting you on that being right + testing looks good.

Let's hope I avoided bikeshedding as much as possible :P


Tested:
- [x] Installing mod
- [x] Updating mod
- [x] Card updates accordingly after installing/updating mod
- [x] Opening Thunderstore page of a mod

Alystrasz:
> With the blurred background, there's now a brighter edge around the edges of the application when in ThunderstoreView, is that intended?

Definitely not, it's another example of inconsistent rendering across different platforms.
I'll check how to prevent this on Windows.

src-vue/src/utils/NorthstarMod.d.ts
```diff
@@ -1,6 +1,6 @@
 // Matches Rust struct (in lib.rs).
 export interface NorthstarMod {
-    name: String,
-    thunderstore_mod_string?: String,
+    name: string,
+    thunderstore_mod_string?: string,
```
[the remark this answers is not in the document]
Alystrasz: `String` is the JavaScript class that allows you to create strings (e.g. `const s = new String("hello there")`), which nobody uses since you don't need it (e.g. `const s = "hello there"`).

On the other hand, `string` is a TypeScript type.

src-vue/src/App.vue
```diff
@@ -29,13 +29,20 @@ export default {
     close() {
       appWindow.close()
     }
-  }
+  },
+    computed: {
+      bgStyle(): string {
+          // @ts-ignore
```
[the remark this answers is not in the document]
Alystrasz: Without this, linter prevents transpiling since it doesn't know `this.$route` member.

src-vue/src/components/ThunderstoreModCard.vue
```diff
@@ -0,0 +1,248 @@
...
+            }
+            if (this.isBeingUpdated) {
+                return ThunderstoreModStatus.BEING_UPDATED;
+            }
+
+            // Ensure mod is up-to-date.
+            const tsModPrefix = this.getThunderstoreDependencyStringPrefix(this.latestVersion.full_name);
+            const matchingMods: NorthstarMod[] = this.$store.state.installed_mods.filter((mod: NorthstarMod) => {
+                if (!mod.thunderstore_mod_string) return false;
+                return this.getThunderstoreDependencyStringPrefix(mod.thunderstore_mod_string!) === tsModPrefix;
```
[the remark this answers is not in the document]
Alystrasz: `NorthstarMod`'s `thunderstore_mod_string` member is typed as such: `string | undefined`, meaning that it can be either a string or undefined.
That's why you can't invoke `method(thunderstore_mod_string)` directly; method takes a string as input, and you're not sure the input is an actual string.
By putting a `!` after your variable, you say that as a developer, you previously ensured that `thunderstore_mod_string` is not undefined, and it can be used as a string.

src-vue/src/views/ThunderstoreModsView.vue
Alystrasz: Yeah, you can import code in many different ways in Javascript :)
Components are not exposed the same way as JavaScript modules though.

src-vue/src/App.vue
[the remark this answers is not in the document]
GeckoEidechse: Should we investigate that in the future or is it basically a non-issue? ^^

src-vue/src/utils/NorthstarMod.d.ts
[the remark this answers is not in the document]
GeckoEidechse: ugh, I swear I'm gonna rewrite frontend in pure webassembly at some point just so I don't have to deal with JS/TS stuff >:)

src-vue/src/App.vue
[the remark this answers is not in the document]
Alystrasz: I think we can leave this as-it :)

src-vue/src/utils/NorthstarMod.d.ts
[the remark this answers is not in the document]
Alystrasz: You don't have a choice for now, unfortunately 😄

Alystrasz:
> I'll check how to prevent this on Windows.

Can you try again?
I tried a fix with 286032dd3705b87f903939f8a17e91f0b1c8584d

GeckoEidechse:
> > I'll check how to prevent this on Windows.
> 
> Can you try again? I tried a fix with [286032d](https://github.com/GeckoEidechse/FlightCore/commit/286032dd3705b87f903939f8a17e91f0b1c8584d)

Yup that seems to work :D

GeckoEidechse (approved):
Blur issue is now fixed on Windows. Overally LGTM both in testing and code ^^

PR #56, opened 2022-11-18
feat: Return Thunderstore mod string of installed mods.

GeckoEidechse:
This updates the returned NorthstarMod struct/object to also contain the related Thunderstore mod that a Northstar mod may belong to.

- [x] Update struct in backend
- [x] Parse Thunderstore mod string and add it to respective mod struct
- [x] Update interface in frontend

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -114,7 +114,7 @@ fn parse_installed_mods(game_install: GameInstall) -> Result<Vec<(String, Option
         let thunderstore_mod_string =
             match parse_mod_json_for_thunderstore_mod_string(mod_json_path.clone()) {
                 Ok(thunderstore_mod_string) => Some(thunderstore_mod_string),
-                Err(err) => None,
+                Err(_err) => None,
             };
 
         mods.push((mod_name, thunderstore_mod_string));
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -55,14 +55,35 @@ fn parse_mod_json_for_mod_name(mod_json_path: String) -> Result<String, anyhow::
     Ok(mod_name.to_string())
 }
 
+/// Parses `mod.json` for Thunderstore mod string
+// TODO: Maybe pass PathBuf or serde json object
+fn parse_mod_json_for_thunderstore_mod_string(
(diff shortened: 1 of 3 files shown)


Alystrasz:
Correct me if I'm not right: this retrieves the Thunderstore string dependency of a mod from its `mod.json`?

src-tauri/src/mod_management/mod.rs
```diff
@@ -90,11 +111,17 @@ fn get_installed_mods(game_install: GameInstall) -> Result<Vec<String>, String>
                 continue;
             }
         };
-        mod_names.push(mod_name);
+        let thunderstore_mod_string =
+            match parse_mod_json_for_thunderstore_mod_string(mod_json_path.clone()) {
+                Ok(thunderstore_mod_string) => Some(thunderstore_mod_string),
+                Err(err) => None,
```
Alystrasz, line 117:
```suggestion
                Err(_) => None,
```
Compiler throws a warning with this unused variable.

GeckoEidechse: Whoops, no clue how I missed that. Thanks for catching it. Fixed with 24eed887c650e3816460166743ee6b4cb17b09e5

Maybe I should look into showing CI warnings on compiler warnings in the future ^^

GeckoEidechse:
> Correct me if I'm not right: this retrieves the Thunderstore string dependency of a mod from its `mod.json`?

Exactly.

> What enforces mods to have a `"ThunderstoreModString"` field?

This is done by FlightCore when installing a mod via it.

https://github.com/GeckoEidechse/FlightCore/blob/38e12489a517662157d85ec6efec00f225cccc9c/src-tauri/src/mod_management/mod.rs#L156-L181

This does mean however that manually installed mods or mods installed via other mod managers will not show up as "Installed"  on #54. There's no of mapping a Northstar mod to a Thunderstore mod without tacking on some additional info to that Northstar mod when installing and modifying `mod.json` is the approach I went with ^^

As for compatibility with other mod managers, VTOL doesn't store state at all and the button will still say "_Install_" after installing a mod. Viper stores  the `manifest.json` inside the mod's folder which I might consider doing in the future as well. (Haven't thought about that before checking, would require a change in `libthermite`).

Alystrasz (approved):
Thanks for the clarification.

PR #59, opened 2022-11-20
feat: Option to launch NS bypassing update check

GeckoEidechse:
Hidden behind dev view. Helps with testing as dev builds of Northstar do not update patch number so an update check bypassing button is necessary to launch dev builds of Northstar.

### DIFF
--- a/src-vue/src/plugins/store.ts
+++ b/src-vue/src/plugins/store.ts
@@ -114,7 +114,26 @@ export const store = createStore<FlightCoreStore>({
                 }
             }
         },
-        async launchGame(state: any) {
+        async launchGame(state: any, no_checks = false) {
+            let game_install = {
+                game_path: state.game_path,
+                install_type: state.install_type
+            } as GameInstall;
+
+            if (no_checks) {
+                await invoke("launch_northstar_caller", { gameInstall: game_install })
+                    .then((message) => {
+                        console.log("Launched with bypassed checks");
+                        console.log(message);
+                    })
+                    .catch((error) => {
+                        console.error(error);
+                        alert(error);
+                    });
+
+                return;
+            }
+
             // TODO update installation if release track was switched
             switch (state.northstar_state) {
                 // Install northstar if it wasn't detected.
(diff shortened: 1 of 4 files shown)


src-vue/src/views/DeveloperView.vue
```diff
@@ -119,6 +122,24 @@ export default defineComponent({
                 position: 'bottom-right'
             });
         },
+        async launchGameWithoutChecks() {
```
Alystrasz, line 125: Why did you duplicate the `launchGame` method?
I would keep the original method, and add a `skipChecks` flag as a parameter to it.

Alystrasz (changes requested):

src-vue/src/views/DeveloperView.vue
Alystrasz: You can simply add a parameter to the commit call
https://vuex.vuejs.org/guide/mutations.html#commit-with-payload

GeckoEidechse: Thanks for the link. Implemented with cced173115613ad68b49036937410578d5bb51ed

src-vue/src/plugins/store.ts
```diff
@@ -127,7 +127,26 @@ export const store = createStore<FlightCoreStore>({
...
+        async launchGame(state: any, no_checks = false) {
+            let game_install = {
+                game_path: state.game_path,
+                install_type: state.install_type
+            } as GameInstall;
+
+            if (no_checks) {
+                await invoke("launch_northstar_caller", { gameInstall: game_install })
+                    .then((message) => {
+                        console.log("Launched with bypassed checks");
```
GeckoEidechse, line 139: Before you complain about leftover console logs again @Alystrasz, I'm thinking maybe we should have one here so that once we actually create log files or something for errors we see if the user actually tried to launch with bypassing checks ^^

Alystrasz:
I thought this was supposed to bypass launch checks?

image

GeckoEidechse:
> I thought this was supposed to bypass launch checks?

It only bypasses the update check. I'd need to update some backend code to bypass those checks as well. Should I?

Alystrasz:
Dunno, I thought you wanted to skip all checks at once :)

GeckoEidechse:
> Dunno, I thought you wanted to skip all checks at once :)

Changed it in 183d9510e4b57cc3b4dbec5cd26efb2d05a738b6 to allow bypassing all checks (the exception being OS check cause the way to launch the game changes per OS).

Note that Northstar won't launch (properly) if Origin isn't running.

Alystrasz (approved):

Alystrasz:
Tested successfully on Windows.

PR #61, opened 2022-11-20
feat: Store game path in persistent store

GeckoEidechse:
WIP commit still the whole logic is a bit clunky, might wanna refactor

### DIFF
--- a/src-vue/src/plugins/store.ts
+++ b/src-vue/src/plugins/store.ts
@@ -105,6 +105,14 @@ export const store = createStore<FlightCoreStore>({
                     }
                     state.install_type = InstallType.UNKNOWN;
 
+                    let game_install = {
+                        game_path: selected,
+                        install_type: InstallType.UNKNOWN
+                    } as GameInstall;
+
+                    // Save change in persistent store
+                    await persistentStore.set('game-install', { value: game_install });
+
                     // Check for Northstar install
                     store.commit('checkNorthstarUpdates');
                 }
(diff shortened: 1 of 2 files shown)


GeckoEidechse:
In particular we probably just wanna store path always, not just when setting custom one. For that I will also need to store type, not just path.

GeckoEidechse:
Storing custom paths should work now. Just not sure if I want to merge this before I have a way of nuking persistent store in the face of errors...

src-vue/src/plugins/store.ts
```diff
@@ -226,21 +234,51 @@ async function _initializeApp(state: any) {
...
+                    duration: 0 // Duration `0` means the notification will not auto-vanish
+                });
             });
-        });
-    const typedResult: GameInstall = result as GameInstall;
-    state.game_path = typedResult.game_path;
-    state.install_type = typedResult.install_type;
+    }
+
+    if (result !== undefined) { // Found some form of value for gameinstall
```
GeckoEidechse, line 271: This check catches an exception that currently happens on `main` when no valid gamepath was found.

Alystrasz:
When opening the app for the first time (i.e. without selecting a path), the button displays "Install" instead of "Select a game folder".

image

src-vue/src/plugins/store.ts
```diff
@@ -226,21 +234,51 @@ async function _initializeApp(state: any) {
...
-            console.error(err);
-            notification_handle = ElNotification({
-                title: 'Titanfall2 not found!',
-                message: "Please manually select install location",
-                type: 'error',
-                position: 'bottom-right',
-                duration: 0 // Duration `0` means the notification will not auto-vanish
+    var result = undefined;
+    var persistent_game_install = (await persistentStore.get('game-install')) as any;
+    console.log(persistent_game_install);
```
Alystrasz, line 239: Don't forget to remove this :)

Alystrasz (changes requested):

GeckoEidechse:
> When opening the app for the first time (i.e. without selecting a path), the button displays "Install" instead of "Select a game folder".

Hmm, weird. Does it have a path set in settings?

GeckoEidechse:
> > When opening the app for the first time (i.e. without selecting a path), the button displays "Install" instead of "Select a game folder".
> 
> Hmm, weird. Does it have a path set in settings?

Ok, can reproduce, checking what's causing it now ^^

GeckoEidechse:
> > > When opening the app for the first time (i.e. without selecting a path), the button displays "Install" instead of "Select a game folder".
> > 
> > 
> > Hmm, weird. Does it have a path set in settings?
> 
> Ok, can reproduce, checking what's causing it now ^^

Fixed with 66101c22128ce81618d209f324f2d56873b21b17

src-vue/src/plugins/store.ts
GeckoEidechse: Removed with aaf142767ea36bdbc87345612e64ea6610999190 ^^

Alystrasz (approved):

PR #63, opened 2022-11-21
refactor: Start moving NS related logic to own module

GeckoEidechse:
Moving a single function initially for PoC and to keep changes small.


GeckoEidechse:
Similarly to a module for dealing with NS mod related stuff, we also want one for dealing with Northstar itself. ^^

Alystrasz (approved):

PR #66, opened 2022-11-23
edit: remove text about Viper showing stack trace

0neGal:
With Viper 1.6.3, stack traces are not shown by default, instead a simple "Unknown Error!" toast is shown, where a user can choose to click it, to then see the stack.

Whether the text should be completely removed or not, is of course, not up to me, I just thought it would be an idea to at the very least, rephrase it, now that stack traces aren't shown by default.


GeckoEidechse (approved):
Fine by me. There's no point criticising issues that have been since fixed :P

PR #67, opened 2022-11-25
refactor: Move function to dedicated module

GeckoEidechse:
Getting log files is part of repair and troubleshooting and should therefore be in the appropriate module.


Alystrasz (approved):
This isn't used anywhere in backend code nor UI?
LGTM anyway.

PR #68, opened 2022-11-25
refactor: Move release channel button to settings

GeckoEidechse:
In a discussion in voice-chat with Emma and Spoon we concluded that switching to Northstar RC release channel should be more accessible without activating DevView.

The goal is to strike a balance between easier to switch to release candidates while also being hard enough to deter someone from accidentally or unknowingly enable it and keeping it enabled.

A suggested idea was having a checkbox to enable "Playtester mode" that then in turns allows switching release channel in PlayView using a dropdown menu or similar.


GeckoEidechse:
Examples from other popular applications for enabling playtester mode / selecting release channel

image
(Note that in the case of Battle.net here it talks about the launcher, while in our case the release channel switching is for Northstar, not FlightCore)

image

dropdown menu

Alystrasz:
I like Battle.net way of doing it, ping me if you need some help on the dropdown UI :)

GeckoEidechse:
> I like Battle.net way of doing it, ping me if you need some help on the dropdown UI :)

Given that it's basically all frontend, I wouldn't mind a helping hand :>

Alystrasz:
Adding some UI with #86

GeckoEidechse:
Superseded by #86

PR #69, opened 2022-11-25
feat: Rebuild `enabledmods.json` if value not found

GeckoEidechse:
When toggling a mod on enabled mods that is not present in `enabledmods.json` we would previously just error out. Now we rebuild the `enabledmods.json` file in a Northstar compatible format and then toggle the mod in question.

This means no more
image

:D

Also, nice PR number :3

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -10,6 +10,33 @@ use app::get_enabled_mods;
 
 use json5;
 
+/// Gets all currently installed and enabled/disabled mods to rebuild `enabledmods.json`
+pub fn rebuild_enabled_mods_json(game_install: GameInstall) -> Result<(), String> {
+    let enabledmods_json_path = format!("{}/R2Northstar/enabledmods.json", game_install.game_path);
+    let mods_and_properties = get_installed_mods_and_properties(game_install)?;
+
+    // Create new mapping
+    let mut my_map = serde_json::Map::new();
+
+    // Build mapping
+    for ns_mod in mods_and_properties.into_iter() {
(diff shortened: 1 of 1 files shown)


src-tauri/src/mod_management/mod.rs
```diff
@@ -10,6 +10,33 @@ use app::get_enabled_mods;
...
+    // Build mapping
+    for ns_mod in mods_and_properties.into_iter() {
+        my_map.insert(ns_mod.name, serde_json::Value::Bool(ns_mod.enabled));
+    }
+
+    // Turn into serde object
+    let obj = serde_json::Value::Object(my_map);
+
+    // Write to file
+    println!("{}", serde_json::to_string_pretty(&obj).unwrap());
```
Alystrasz, line 30: Unnecessary print here

Alystrasz (changes requested):
Tested on Windows, works fine when a mod key is missing.
This however doesn't work when:
* `enabledmods.json` file does not exist
* `enabledmods.json` does not contain a valid JSON document (is empty, for instance)

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Removed with 5f2c333b4cf88f40bc02c8bf091ad9d579f74b0a

GeckoEidechse:
> This however doesn't work when:
> 
> * `enabledmods.json` file does not exist
> * `enabledmods.json` does not contain a valid JSON document (is empty, for instance)

Didn't even think about that, good catch :D

Fixed in acdf8107bd6c5824ee6eab6bfd123a9a07be5c71

Alystrasz (approved):
Tested successfully on Linux with above-mentioned use-cases.

PR #72, opened 2022-11-26
feat: RepairView

GeckoEidechse:
This view will be used for various Northstar and FlightCore related repair tasks.


src-vue/src/views/SettingsView.vue
```diff
@@ -12,6 +12,13 @@
                 <el-button icon="Folder" @click="updateGamePath"/>
             </template>
         </el-input>
+
+        <h3>Other:</h3>
+
+        <el-button type="primary" @click="openRepairView">
+            Open Repair options
+        </el-button>
```
GeckoEidechse, line 20: Not sure what to call the section the button is in so I went with "_Other_" for now ^^

Rendered:

image

src-vue/src/views/RepairView.vue
```diff
@@ -0,0 +1,108 @@
...
+    <div class="fc_repair__container">
+        <h2>Repair:</h2>
+
+        <h3>Northstar:</h3>
+
+        <el-button type="primary" @click="disableAllModsButCore">
+            Disable all but core mods
+        </el-button>
+
+        <h3>FlightCore:</h3>
```
GeckoEidechse, line 11: Both section headers ("_Northstar_" and "_FlightCore_") are `<h3>` but they render differently for some reason. No idea why :c

image

src-vue/src/plugins/store.ts
```diff
@@ -43,6 +45,7 @@ export const store = createStore<FlightCoreStore>({
     state(): FlightCoreStore {
         return {
             developer_mode: false,
+            repair_view_visible: false,
```
GeckoEidechse, line 48: Should building in _debug mode_ (not to be confused with _dev mode_ which is activated by clicking version number 5 times), also show RepairView in menu bar by default like it does for DevView?

src-vue/src/views/RepairView.vue
```diff
@@ -0,0 +1,108 @@
+<template>
+    <div class="fc_repair__container">
+        <h2>Repair:</h2>
```
GeckoEidechse, line 3: Add a button to close RepairView and remove from menu bar (like we have for DevView)?

GeckoEidechse (commented):
Comments below about stuff I'm unsure about or that we might want to address ^^

Alystrasz:
We already talked about this, a menu should keep same amount of items during execution, the only tolerable option being the "developer" view.
I suggest you open this new view in a modal dialog or in a new window.

GeckoEidechse:
Closing in favour of #129

PR #74, opened 2022-11-26
refactor: Define core mods in global const

GeckoEidechse:
This way we only define them in one place which makes adding/removing core mods (e.g. `Northstar.Coop`) easier in the future.

### DIFF
--- a/src-tauri/src/northstar/mod.rs
+++ b/src-tauri/src/northstar/mod.rs
@@ -1,6 +1,12 @@
 //! This module deals with handling things around Northstar such as
 //! - getting version number
 
+pub const CORE_MODS: [&str; 3] = [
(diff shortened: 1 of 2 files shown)


src-tauri/src/northstar/mod.rs
```diff
@@ -1,6 +1,12 @@
 //! This module deals with handling things around Northstar such as
 //! - getting version number
 
+pub const CORE_MODS: [&str; 3] = [
```
Alystrasz, line 4: I suppose `3` is the length of the array?

GeckoEidechse: Yup. Global consts require their type to be explicitly defined in Rust. So basically `str` array of size 3 ^^

PR #71, opened 2022-11-26
docs: Improve Readme structure to be more end-user friendly and structured

GeckoEidechse:
- [x] For end-users we need an easy way to 1-click download FlightCore (e.g. download buttons in README) --> #87 
- [x] Have developer docs in docs subfolder and/or in frontend/backend specific folder --> #75
- [x] Improve developer docs -->  #88
- [x] Cleanup -> #89
- [x] ???

PR #76, opened 2022-11-26
feat: Add a "useless" button to teach backend call

GeckoEidechse:
The idea is to have some logic that adds as an easy example for a new developer on how to call backend from frontend to learn Tauri by example


GeckoEidechse:
Sorta part of #71

Alystrasz (changes requested):
I don't see the point of adding useless code, since developers can take example on already-existing code.

GeckoEidechse:
> I don't see the point of adding useless code, since developers can take example on already-existing code.

Yeah [we have it in docs now anyway](https://github.com/R2NorthstarTools/FlightCore/blob/main/docs/DEVELOPMENT.md#interacting-between-frontend-and-backend) so I don't think we need it anymore.

That being said, I would see value in adding a "useless" button that calls an empty Rust function, so that for backend development one can easily use this to call their own backend code.

At least a thing I noticed often while implementing new backend functionality is that I spend at least 5-10 minutes adding  frontend code for calling some backend function, just to remove the code again for PR/merge once properly implemented and hooked up.

As such there would be value in having a button that by default is "useless" but can be quickly hooked up to some Rust function during development.

Would we still get PRs where someone forgets to restore the "useless" button back to its original state?
-> Definitely

Will we catch it in PR review and block merge until cleaned?
-> Definitely :P

Alystrasz:
> That being said, I would see value in adding a "useless" button that calls an empty Rust function, so that for backend development one can easily use this to call their own backend code.

I'm now begin to understaand why you created this PR, but I know you're a good dev, I don't think you spend 10 minutes to add a UI button each time you want to test a backend feature 😃 

And as you've said:

> Would we still get PRs where someone forgets to restore the "useless" button back to its original state?
-> Definitely

I'm not sure we would catch it each time this happens!

GeckoEidechse:
> > That being said, I would see value in adding a "useless" button that calls an empty Rust function, so that for backend development one can easily use this to call their own backend code.
> 
> I'm now begin to understaand why you created this PR, but I know you're a good dev, I don't think you spend 10 minutes to add a UI button each time you want to test a backend feature :smiley:

10 minutes may have been a bit of an exaggeration but I'm still annoyed every time I have to add a button just to test a backend function ^^"

> And as you've said:
> 
> > Would we still get PRs where someone forgets to restore the "useless" button back to its original state?
> > -> Definitely
> 
> I'm not sure we would catch it each time this happens!

I would :>

GeckoEidechse:
> > I'm not sure we would catch it each time this happens!
> 
> I would :>

Actually I could probably write a test for that. Wanted to write testing infrastructure anyway ^^

GeckoEidechse:
I'm closing this for now until I have a better solution ^^

catornot:
if you want a solution to the changes to this feature being commit in random prs you could make a proc macro that only panics if it detects changes and if it is being compiled by ci.

GeckoEidechse:
That sounds super overkill xD

catornot:
then what is the main issue here?

PR #77, opened 2022-11-27
chore: Update Tauri depedencies to 1.2

GeckoEidechse:
Update Tauri dependencies to `1.2`
Also requires bumping Rust to `1.62`


Alystrasz (approved):
Successfully tested on Windows.

PR #78, opened 2022-11-27
docs: Introduce languages used in FlightCore

GeckoEidechse:
Introduces languages and tools used in FlightCore


Alystrasz (approved):
Why make several PRs for the same documentation topic though x)

GeckoEidechse:
> Why make several PRs for the same documentation topic though x)

~~So requested change in one section doesn't block another~~ To bump my commit stats :>

PR #79, opened 2022-11-27
docs: Link Tauri intro and tutorial

GeckoEidechse:
To help with on-boarding new contributors.


Alystrasz (changes requested):
Before putting Tauri resources links in this document, you should explain what it is, and why you're talking about it here.

Alystrasz (approved):
Tauri introduction was added by #78

PR #81, opened 2022-11-28
docs: Mention design goals of FlightCore

GeckoEidechse:
Mention design goals of FlightCore in developer instructions.


Alystrasz (approved):
Shouldn't you be following classes at the uni? :smile:

PR #84, opened 2022-11-28
style: Auto-format all Rust code

GeckoEidechse:
Just manually ran the auto-format over all the Rust code ^^


Alystrasz (approved):
You should do this more often ;)

PR #85, opened 2022-11-28
refactor: Move variable to global const

GeckoEidechse:
Makes it easier to update/re-use in the future.


Alystrasz (dismissed):
This list of mods to ignore is already present UI-side, so users cannot already install those: https://github.com/R2NorthstarTools/FlightCore/blob/main/src-vue/src/plugins/store.ts#L236
Can we find a way to avoid duplication here?

GeckoEidechse:
> This list of mods to ignore is already present UI-side, so users cannot already install those: https://github.com/R2NorthstarTools/FlightCore/blob/main/src-vue/src/plugins/store.ts#L236
> Can we find a way to avoid duplication here?

We still need to do the check in the backend cause mods can specify Northstar as dependency for example and when going over dependencies recursively we ofc course do not want to install Northstar as a mod dependency. ^^

So yeah need a second check in backend the same way you handle untrusted user input on the backend on a webapp.

I do however agree, that we should have a single place for specifying them so that we only need to update the list in one place when we want to change something. ^^

Alystrasz:
I couldn't build due to compiler error:
```rust
error[E0658]: use of unstable library feature 'core_ffi_c'
  --> /home/remy/.cargo/registry/src/github.com-1ecc6299db9ec823/zstd-sys-2.0.3+zstd.1.5.2/src/bindings_zdict_std.rs:84:59
   |
84 |     pub fn ZDICT_getErrorName(errorCode: usize) -> *const ::core::ffi::c_char;
   |                                                           ^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #94501 <https://github.com/rust-lang/rust/issues/94501> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `zstd-sys` due to 31 previous errors
```
and I've no idea where it's coming from.

GeckoEidechse:
> I couldn't build due to compiler error:
> 
> ```rust
> error[E0658]: use of unstable library feature 'core_ffi_c'
>   --> /home/remy/.cargo/registry/src/github.com-1ecc6299db9ec823/zstd-sys-2.0.3+zstd.1.5.2/src/bindings_zdict_std.rs:84:59
>    |
> 84 |     pub fn ZDICT_getErrorName(errorCode: usize) -> *const ::core::ffi::c_char;
>    |                                                           ^^^^^^^^^^^^^^^^^^^
>    |
>    = note: see issue #94501 <https://github.com/rust-lang/rust/issues/94501> for more information
> 
> For more information about this error, try `rustc --explain E0658`.
> error: could not compile `zstd-sys` due to 31 previous errors
> ```
> 
> and I've no idea where it's coming from.

I had this as well I while back. It's caused by a library using newer functionality.

When testing I found `1.62` of Rust to have it enabled in stable, hence the Rust version bump in https://github.com/R2NorthstarTools/FlightCore/pull/77, but [checking docs now](https://doc.rust-lang.org/stable/core/ffi/type.c_char.html), it seems to be only available in stable since `1.64`. Not sure why this PR triggers the compile error. I'll make a PR bumping min Rust version.

I also assumed it would always compile with the version number specified in `Cargo.toml` but that seems not to be the case hence I cannot even reproduce it locally.

Installed rust version on my end

```
$ rustc --version
rustc 1.65.0 (897e37553 2022-11-02)
```

Alystrasz (approved):
Confirmed issue was coming from my outdated rust install.

Alystrasz (approved):
Confirmed issue was coming from my outdated rust install.

PR #86, opened 2022-11-28
feat: Channel selector

Alystrasz:
Adds a version selector to the main screen, which allows users to switch between stable and release candidate channels.

image


src-vue/src/components/PlayButton.vue
```diff
@@ -39,6 +49,28 @@ export default defineComponent({
                     value: Object.keys(ReleaseCanal)[Object.values(ReleaseCanal).indexOf(v)]
                 }
             });
+        },
+        selectOptions(): {label: string, options: {value: ReleaseCanal, label: string}[]}[] {
+            return [
+                {
+                    label: 'In development',
```
GeckoEidechse, line 56: I'm considering adding a _pull-launcher-and-mods-from-main_ in the future which is closer to "In development". So maybe we should call this something that indicates that it's a release candidate that has been tested to some minimum degree. Not sure how obvious "Release Candidate"  is to players... Maybe "Beta"?

It should still indicate though that it's not considered "stable".

src-vue/src/components/PlayButton.vue
```diff
@@ -53,17 +85,53 @@ export default defineComponent({
...
+            :label="group.label"
+        >
+            <el-option
+                v-for="item in group.options"
+                :key="item.value"
+                :label="item.label"
+                :value="item.value"
+            />
+        </el-option-group>
+    </el-select>
```
GeckoEidechse, line 101: Maybe we should hook this up to a button in settings so by default it's hidden but clicking "Enable testing release channels" will cause the button to show. Just so that someone just starting out on Northstar doesn't end up on release candidate by accident ^^

And then we should also persist to disk if the dropdown button to switch release channel was pressed.
(I can do the persisting to disk part if you want to)

GeckoEidechse (changes requested):
Looks awesome :D

Main feedback is wording and availability of the dropdown menu, i.e. it should not be seen by default but requires pressing a button to increase friction a bit to prevent someone unknowingly enabling it ^^

src-vue/src/components/PlayButton.vue
Alystrasz: Should I replace the "Toggle release candidate" button by this "Enable testing release channels" button?

GeckoEidechse: Maybe a checkbox instead of a button so that the state on/off is clear but yeah, something like that would be great ^^

src-vue/src/components/PlayButton.vue
```diff
@@ -39,6 +49,28 @@ export default defineComponent({
...
+                    label: 'In development',
+                    options: [
+                        {
+                            value: ReleaseCanal.RELEASE_CANDIDATE,
+                            label: 'Northstar development release',
+                        },
+                    ]
+                },
+                {
+                    label: 'Live',
```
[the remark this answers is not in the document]
Alystrasz: Fixed in 66795215af6abd275a926859e7702224b32ff925

src-vue/src/components/PlayButton.vue
Alystrasz: Fixed in 66795215af6abd275a926859e7702224b32ff925

src-vue/src/components/PlayButton.vue
Alystrasz: Added in ff915f21d9a69caf02844aee997810b8e275aa63

src-vue/src/views/SettingsView.vue
```diff
@@ -46,6 +46,15 @@ export default defineComponent({
...
             return this.$store.state.flightcore_version;
         },
+        enableReleasesSwitch: {
+            get(): boolean {
+                return this.$store.state.enableReleasesSwitch;
+            },
+            set(value: boolean): void {
+                this.$store.state.enableReleasesSwitch = value;
+                persistentStore.set('northstar-releases-switching', { value });
+            }
```
GeckoEidechse, line 56: When disabling showing other release channel, we also wanna switch back to stable release.

Otherwise one can enable release switching -> switch to release candidate -> disable release switching -> they are now "stuck" on release candidate.

GeckoEidechse: We could delay fixing that to post merge though if you wanna ;)

GeckoEidechse (changes requested):
Code looks good (as far as I can tell). Found two issues in testing. Once resolved (or at least one of them), should be ready to merge ^^

src-vue/src/components/PlayButton.vue
```diff
@@ -39,6 +49,40 @@ export default defineComponent({
...
             });
+        },
+        selectOptions(): {label: string, options: {value: ReleaseCanal, label: string}[]}[] {
+            return [
+                {
+                    label: 'Beta',
+                    options: [
+                        {
+                            value: ReleaseCanal.RELEASE_CANDIDATE,
+                            label: 'Northstar development release',
```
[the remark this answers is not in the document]
Alystrasz: Addressed in efa35876365e19187f00a864bd183bf4ef3b868f

GeckoEidechse (approved):
Requested changes confirmed working in testing :D

PR #90, opened 2022-11-29
docs: Mention ways of reaching out

GeckoEidechse:
Just mention some way how to ask for help when trying to contribute.


GeckoEidechse:
@Alystrasz should I add you as a point of contact as well? ^^

Alystrasz:
Yep you can :)

GeckoEidechse:
> Yep you can :)

Just add `@Akystrasz` or sth else?

Alystrasz:
`@Alystrasz` is better 😄

PR #91, opened 2022-11-29
feat: Add scrollbar to all views

### LINKED ISSUE #65: Enable custom scrollbar on all views
ChangelogView and ModsView use custom scrollbar for overflow content. The other views do not and as such the whole window becomes scrollable if the content overflows. Example:

image

Obviously this shouldn't happen but also just trying to e.g. wrapping DevView inner container in `<el-scrollbar>` does nothing either. So uh @Alystrasz help please? ^^"

Alystrasz:
I hope title is self-explanatory :)
I took advantage of this to remove some duplication in CSS styles.

Closes #65.

### DIFF
--- a/src-vue/src/style.css
+++ b/src-vue/src/style.css
@@ -34,3 +34,9 @@ body {
   --el-scrollbar-opacity: 0.5;
   --el-scrollbar-hover-opacity: 0.7;
 }
+
+.fc-container {
+    position: relative;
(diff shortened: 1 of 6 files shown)


GeckoEidechse (approved):
Works fine testing :D

Code looks fine as far as I can tell (which does not say a lot tbh :P)

PR #94, opened 2022-11-30
refactor: Remove the "Origin running" check from frontend

GeckoEidechse:
We perform the check twice (both frontend and backend), however the frontend doesn't check for OS, meaning on Linux it will always error out with "Origin not running".

So instead we should just do the check only in backend and propagate the right error message to frontend.


Alystrasz (approved):
Tested successfully on Windows.

PR #95, opened 2022-11-30
fix: Do not crash on Origin install find failure

GeckoEidechse:
The unwraps would cause the thread to crash in case FlightCore is unable to use Powershell output to find Origin Titanfall2 install path. Instead we just return the error message which is caught and handled by the parent function.


GeckoEidechse:
Compiler error, have to check when on Windows host :c

GeckoEidechse:
Fixed the compiler error ^^

GeckoEidechse:
Superseded by #99

PR #98, opened 2022-11-30
feat: Threshold before showing update notification

GeckoEidechse:
This way we don't spam the end-user with an update notification before an update is fully built and released.

### DIFF
--- a/src-tauri/src/github/release_notes.rs
+++ b/src-tauri/src/github/release_notes.rs
@@ -52,7 +52,38 @@ pub async fn check_is_flightcore_outdated() -> Result<bool, String> {
     let version = format!("v{}", version);
 
     // TODO: This shouldn't be a string compare but promper semver compare
-    Ok(version != newest_release_version)
+    let is_outdated = version != newest_release_version;
+
+    // If outdated, check how new the update is
+    if is_outdated {
+        // Extract release date from JSON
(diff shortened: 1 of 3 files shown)


GeckoEidechse:
Wtf? It ran locally just fine :(

src-tauri/src/github/release_notes.rs
```diff
@@ -52,7 +52,38 @@ pub async fn check_is_flightcore_outdated() -> Result<bool, String> {
...
+    // If outdated, check how new the update is
+    if is_outdated {
+        // Extract release date from JSON
+        let release_date = json_response
+            .get("published_at")
+            .and_then(|value| value.as_str())
+            .unwrap();
+
+        // Time to wait (1 day)  h *  m *  s
+        let threshold_seconds = 24 * 60 * 60;
```
Alystrasz, line 66: So if I get it right, users have to wait an entire day from the moment a new release is published?

GeckoEidechse: Yup

GeckoEidechse: Adjusted to 2 hours now with 8f255f341f4e067f98dae85dd78006c2fe8df86b

Alystrasz (approved):

PR #99, opened 2022-12-01
fix: Stop opening PS to get Origin game path

GeckoEidechse:
Now using a library instead of calling a PowerShell command should prevent the PowerShell window from opening.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -30,6 +30,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
+dependencies = [
+ "getrandom 0.2.8",
+ "once_cell",
+ "version_check",
+]
+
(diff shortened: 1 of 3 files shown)


GeckoEidechse:
Supersedes #95

Alystrasz:
How did you test this branch?

GeckoEidechse:
> How did you test this branch?

On Windows I deleted saved path, commented out the [Steam game install path detection code](https://github.com/R2NorthstarTools/FlightCore/blob/2e47666762afc665d8f44e57c4f86ed72d9fcc13/src-tauri/src/lib.rs#L81-L98) so that it would only run the Origin detection and then I checked that it picks the Origin install.

To test the PowerShell window issue one needs to build the app (with updater disabled to prevent build failure due to missing keys), make sure that no persistent game install path is saved and then run it and see if the powershell window opens (haven't tested that part yet).

GeckoEidechse:
> To test the PowerShell window issue one needs to build the app (with updater disabled to prevent build failure due to missing keys), make sure that no persistent game install path is saved and then run it and see if the powershell window opens (haven't tested that part yet).

Tested on my own machine now and got no PowerShell window pop-up.

Alystrasz (approved):
Shell doesn't show up with release build on Windows.

PR #103, opened 2022-12-02
fix: Make sure all non-core mods get disabled

GeckoEidechse:
For this we first rebuild the `enabledmods.json` file. This ensures that all installed mods are actually listed there.

The function was originally created before we could rebuild `enabledmods.json` by checking installed mods.


GeckoEidechse:
For testing, simply remove some entries in `enabledmods.json` and then use the "_Disable all but core mods_" button from the DevView, afterwards check ModsView or launch Northstar. Before this fix, mods not present in `enabledmods.json` would stay enabled. Now they should also get disabled.

PR #105, opened 2022-12-04
feat: Show warning banner on DevView

GeckoEidechse:
Rendered:

image

### DIFF
--- a/src-vue/src/views/DeveloperView.vue
+++ b/src-vue/src/views/DeveloperView.vue
@@ -1,7 +1,7 @@
 <template>
     <div class="fc-container">
         <el-scrollbar>
-            <el-alert title="Warning" type="warning" :closable="false" effect="dark" show-icon>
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Doesn't look the prettiest, I'm happy to take any suggestion @Alystrasz :>

Alystrasz (changes requested):
Red is for errors, you should use some orange :)
And add the warning icon: https://element-plus.org/en-US/component/alert.html#with-icon-and-description

GeckoEidechse:
> Red is for errors, you should use some orange :) 

Tried orange first but it just blends with the background, making it look like an info box, not a warning ^^"

Alystrasz:
Use a brighter orange then 😄

GeckoEidechse:
Updated based on suggestion

Rendered:
image

Alystrasz (approved):
Confirmed working on Linux.

PR #106, opened 2022-12-05
chore: Bump Rust version to 1.65

GeckoEidechse:
Should resolve https://github.com/R2NorthstarTools/FlightCore/pull/85#issuecomment-1336846037

Technically `1.64` would be enough to solve that issue but[ `1.65` adds let-else statements to stable which allow for more compact and readable code in some cases](https://www.youtube.com/watch?v=U-5_bumwH9w). ^^

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -7,7 +7,7 @@ license = ""
 repository = ""
 default-run = "app"
 edition = "2021"
-rust-version = "1.62"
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
@Alystrasz don't forget to `rustup update` in case you get local compile errors due to not satisfying minimum rustc version ^^

Alystrasz (approved):
Confirmed working on Linux.

PR #107, opened 2022-12-05
chore: Bump Node dependencies

GeckoEidechse:
The vscode Vue Language Features extension (Volar) was complaining that the Vue dependency is out of date. Decided to just bump all of them in the process :P


Alystrasz (approved):
Confirmed working on Linux.

PR #108, opened 2022-12-05
chore: Bump Rust dependencies

GeckoEidechse:
If we're already in the process of bumping dependencies, why not these as well ¯\_(ツ)_/¯


Alystrasz (approved):
Confirmed working on Linux.

PR #109, opened 2022-12-06
feat: Expose installed NS mod directory

GeckoEidechse:
This allows other functions to get a mod directory directly which is useful for e.g. deleting a mod.

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -128,13 +128,13 @@ fn parse_mod_json_for_thunderstore_mod_string(
 /// Parse `mods` folder for installed mods.
 fn parse_installed_mods(
     game_install: GameInstall,
-) -> Result<Vec<(String, Option<String>)>, String> {
+) -> Result<Vec<(String, Option<String>, String)>, String> {
     let ns_mods_folder = format!("{}/R2Northstar/mods/", game_install.game_path);
 
     let paths = std::fs::read_dir(ns_mods_folder).unwrap();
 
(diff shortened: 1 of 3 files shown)


GeckoEidechse:
f14ea9b82fd57c355b6957a8280f8025ee85658b additionally makes it easier to add more fields to return in the future.

src-tauri/src/mod_management/mod.rs
```diff
@@ -167,8 +165,17 @@ fn parse_installed_mods(
                 Ok(thunderstore_mod_string) => Some(thunderstore_mod_string),
                 Err(_err) => None,
             };
+        // Get directory path
+        let mod_directory = directory.to_str().unwrap().to_string();
+
+        let ns_mod = NorthstarMod {
+            name: mod_name,
+            thunderstore_mod_string: thunderstore_mod_string,
+            enabled: false, // Placeholder
```
Alystrasz, line 174: Since Northstar considers mods as enabled by default, shouldn't this be true?

GeckoEidechse: Shouldn't matter cause this function is only ever called by [`get_installed_mods_and_properties`](https://github.com/R2NorthstarTools/FlightCore/blob/f14ea9b82fd57c355b6957a8280f8025ee85658b/src-tauri/src/mod_management/mod.rs#L188) where the value is overwritten. ^^

Also I'm planning to combine the two functions (`get_installed_mods_and_properties` and `parse_installed_mods`) into a single one anyway after this is (+ #110 and #111) merged.

Alystrasz (approved):
Confirmed working on Linux.

PR #110, opened 2022-12-06
feat: Add button to delete Northstar mod

GeckoEidechse:
Allows deleting a Northstar mod from within FlightCore

- [x] Requires #109
- [x] https://github.com/R2NorthstarTools/FlightCore/pull/110#pullrequestreview-1224588360 (81f5e7373bc399c646ff29c0fb8a3c647e2f56c4)

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -10,6 +10,8 @@ use app::GameInstall;
 
 use json5;
 
+use crate::northstar::CORE_MODS;
+
 pub const BLACKLISTED_MODS: [&str; 3] = [
     "northstar-Northstar",
     "northstar-NorthstarReleaseCandidate",
@@ -382,4 +384,47 @@ pub async fn fc_download_mod_and_install(
     std::fs::remove_file(path).unwrap();
 
     Ok(())
+}
+
+/// Deletes a given Northstar mod folder
+fn delete_mod_folder(ns_mod_directory: String) -> Result<(), String> {
+    // TODO maybe remove again
+    let ns_mod_dir_path = std::path::Path::new(&ns_mod_directory);
+
+    // Safety check: Check whether `mod.json` exists and exit early if not
+    // If it does not exist, we might not be dealing with a Northstar mod
+    let mod_json_path = ns_mod_dir_path.join("mod.json");
+    if !mod_json_path.exists() {
+        // If it doesn't exist, return an error
+        return Err(format!("mod.json does not exist in {}", ns_mod_directory));
+    }
+
+    match std::fs::remove_dir_all(&ns_mod_directory) {
+        Ok(()) => Ok(()),
+        Err(err) => Err(format!("Failed deleting mod: {err}")),
+    }
+}
+
+/// Deletes a Northstar mod based on its name
+pub fn delete_northstar_mod(game_install: GameInstall, nsmod_name: String) -> Result<(), String> {
+    // Prevent deleting core mod
+    for core_mod in CORE_MODS {
+        if nsmod_name == core_mod {
+            return Err(format!("Cannot remove core mod {nsmod_name}"));
+        }
+    }
+
+    // Get installed mods
+    let installed_ns_mods = get_installed_mods_and_properties(game_install)?;
+
+    // Get folder name based on northstarmods
+    for installed_ns_mod in installed_ns_mods {
+        // Installed mod matches specified mod
+        if installed_ns_mod.name == nsmod_name {
+            // Delete folder
+            return delete_mod_folder(installed_ns_mod.directory);
+        }
+    }
+
+    Err(format!("Mod {nsmod_name} not found to be installed"))
 }
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -388,7 +388,6 @@ pub async fn fc_download_mod_and_install(
 
 /// Deletes a given Northstar mod folder
 fn delete_mod_folder(ns_mod_directory: String) -> Result<(), String> {
-    // TODO maybe remove again
     let ns_mod_dir_path = std::path::Path::new(&ns_mod_directory);
 
     // Safety check: Check whether `mod.json` exists and exit early if not
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -405,6 +405,7 @@ fn delete_mod_folder(ns_mod_directory: String) -> Result<(), String> {
 }
 
 /// Deletes a Northstar mod based on its name
+#[tauri::command]
 pub fn delete_northstar_mod(game_install: GameInstall, nsmod_name: String) -> Result<(), String> {
     // Prevent deleting core mod
     for core_mod in CORE_MODS {
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -128,13 +128,13 @@ fn parse_mod_json_for_thunderstore_mod_string(
 /// Parse `mods` folder for installed mods.
 fn parse_installed_mods(
     game_install: GameInstall,
-) -> Result<Vec<(String, Option<String>)>, String> {
+) -> Result<Vec<(String, Option<String>, String)>, String> {
     let ns_mods_folder = format!("{}/R2Northstar/mods/", game_install.game_path);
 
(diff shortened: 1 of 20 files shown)


src-vue/src/views/ModsView.vue
```diff
@@ -4,9 +4,11 @@
...
             <div>
                 <el-card shadow="hover" v-for="mod in $store.state.installed_mods">
-                    <el-switch style="--el-switch-on-color: #13ce66; --el-switch-off-color: #8957e5" v-model="mod.enabled"
-                        :before-change="() => updateWhichModsEnabled(mod)" :loading="global_load_indicator" />
-                    {{mod.name}}
+                    <el-switch style="--el-switch-on-color: #13ce66; --el-switch-off-color: #8957e5"
+                        v-model="mod.enabled" :before-change="() => updateWhichModsEnabled(mod)"
+                        :loading="global_load_indicator" />
+                    <el-button type="danger" @click="deleteMod(mod)">Delete</el-button>
+                    {{ mod.name }}
```
GeckoEidechse, line 11: Rendered it looks like this which could definitely be improved :P

image

Alystrasz: I assume this will done after this get merged? 😄

Alystrasz: I would like users not to delete their mods accidentally, so we should ask for their confirmation before effectively deleting mods: https://element-plus.org/en-US/component/popconfirm.html

src-tauri/src/mod_management/mod.rs
```diff
@@ -383,3 +385,46 @@ pub async fn fc_download_mod_and_install(
 
     Ok(())
 }
+
+/// Deletes a given Northstar mod folder
+fn delete_mod_folder(ns_mod_directory: String) -> Result<(), String> {
+    // TODO maybe remove again
```
Alystrasz, line 391: Still TODO?

src-tauri/src/mod_management/mod.rs
```diff
@@ -383,3 +385,46 @@ pub async fn fc_download_mod_and_install(
...
+
+    match std::fs::remove_dir_all(&ns_mod_directory) {
+        Ok(()) => Ok(()),
+        Err(err) => Err(format!("Failed deleting mod: {err}")),
+    }
+}
+
+/// Deletes a Northstar mod based on its name
+pub fn delete_northstar_mod(game_install: GameInstall, nsmod_name: String) -> Result<(), String> {
+    // Prevent deleting core mod
```
Alystrasz, line 410: Additionally to throwing an error when trying to remove a core mod, the "delete" button should not appear on core mod tiles in UI.

src-vue/src/views/ModsView.vue
```diff
@@ -63,6 +65,31 @@ export default defineComponent({
 
             this.global_load_indicator = false;
             return true;
+        },
+        async deleteMod(mod: NorthstarMod) {
+            console.log(mod)
```
Alystrasz, line 70: Please remove this :)

Alystrasz (changes requested):
I also would like users not to delete their mods accidentally, so we should ask for their confirmation before effectively deleting mods: https://element-plus.org/en-US/component/popconfirm.html

src-vue/src/views/ModsView.vue
GeckoEidechse: Done with 6a974ac2eb36bf11913b5ea08fec4dcfd2e7f35e

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Whoops, nope. Removed with ad9e055bfa5689c323a4b24667364215a2e0edc0

src-vue/src/views/ModsView.vue
GeckoEidechse: I guess. Unless you want to tack it on to this PR? :eyes:

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Hmm, for that we need to signal core mods from backend to front-end. I don't think we have something for that yet... :/

Alystrasz: I thought we had a list of core mods somewhere? Only in backend maybe?

src-vue/src/views/ModsView.vue
Alystrasz: I would like to reuse the card layout from Thunderstore view, but I still gotta figure out how :)

GeckoEidechse: Yeah, we might have to do some UI prototyping for ModsView ^^

Alystrasz: Kinda out of scope for this then

GeckoEidechse:
> I also would like users not to delete their mods accidentally, so we should ask for their confirmation before effectively deleting mods: https://element-plus.org/en-US/component/popconfirm.html

So I tried doing like

```html
<el-popconfirm title="Are you sure to delete this?">
    <template #reference>
        <el-button type="danger" @click="deleteMod(mod)">Delete</el-button>
    </template>
</el-popconfirm>
```

just like the example in their docs but that still runs the delete function on the initial click instead of the confirmation click. Any idea what I might be doing wrong? ^^"

Alystrasz:
> just like the example in their docs but that still runs the delete function on the initial click instead of the confirmation click. Any idea what I might be doing wrong? ^^"

Yeah, I think you need to move the `deleteMod` call on the popconfirm component, using the `confirm` event
(https://element-plus.org/en-US/component/popconfirm.html#events)

GeckoEidechse:
> > just like the example in their docs but that still runs the delete function on the initial click instead of the confirmation click. Any idea what I might be doing wrong? ^^"
> 
> Yeah, I think you need to move the `deleteMod` call on the popconfirm component, using the `confirm` event (https://element-plus.org/en-US/component/popconfirm.html#events)


Missed that whoops. Yeah, that works :D

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Moved to #120

src-vue/src/views/ModsView.vue
```diff
@@ -69,6 +77,30 @@ export default defineComponent({
...
+                })
+                .catch((error) => {
+                    ElNotification({
+                        title: 'Error',
+                        message: error,
+                        type: 'error',
+                        position: 'bottom-right'
+                    });
+                });
+            this.$store.commit('loadInstalledMods');
```
Alystrasz, line 103: Since `invoke` call is asynchronous, `$store.commit` can be invoked before `.then` or `.catch`; it should be in the `.finally` callback (as I suppose you want to reload mods list after mod deletion).

Alystrasz (changes requested):
Apart from this comment, this PR looks good to me (works on Windows).

src-vue/src/views/ModsView.vue
GeckoEidechse: Uh could you give me a short example how exactly it should look like? Not sure I fully understand ^^"

Alystrasz:
```javascript
await promise("blabla")
    .then(() => {
        // get executed if promise went well
    })
    .catch((err) => {
        // get executed if promised threw an error
    })
    .finally(() => {
        // get executed no matter promise result
    });
```

https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

GeckoEidechse: Oh wait, so you just meant:

```typescript
await invoke("delete_northstar_mod", { gameInstall: game_install, nsmodName: mod.name })
    .then((message) => {
        // Just a visual indicator that it worked
        ElNotification({
            title: `Success deleting ${mod.name}`,
            type: 'success',
            position: 'bottom-right'
        });
    })
    .catch((error) => {
        ElNotification({
            title: 'Error',
            message: error,
            type: 'error',
            position: 'bottom-right'
        });
    })
    .finally(() => {
        this.$store.commit('loadInstalledMods');
    });
```
?

Alystrasz: Yup :)

GeckoEidechse: I thought the `await` should prevent re-ordering but I guess not? Anyway, changed in d0a8c249774260934e65c766d1ac34205ffc93b2 ^^

Alystrasz (approved):
LGTM.

Issue #20, opened 2022-10-15
Launch Northstar "detached"


GeckoEidechse:
Right now Northstar is launched as a child which means it captures all the child's stdout which breaks stuff such as @ASpoonPlaysGames's colour console logs PR.

GeckoEidechse:
Possible useful link for later: https://doc.rust-lang.org/std/process/struct.Stdio.html#method.piped

Alystrasz:
Closed by #112.

PR #112, opened 2022-12-08
feat: Start Northstar properly detached

GeckoEidechse:
Simply run cmd with `start "" "PATH/TO/NORTHSTAR_EXE"`

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -290,11 +290,11 @@ pub fn launch_northstar(
             || matches!(game_install.install_type, InstallType::ORIGIN)
             || matches!(game_install.install_type, InstallType::UNKNOWN))
     {
-        let _output =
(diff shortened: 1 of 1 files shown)


Alystrasz:
Closes #20.
Can you provide a test case which highlights this feature?

Alystrasz:
On Windows 10 with `main`, no console appears at all; is it a big deal? Does it only appear with release builds?
On the other hand, I confirm using this branch properly shows content in a console window.

GeckoEidechse:
> On Windows 10 with `main`, no console appears at all; is it a big deal? Does it only appear with release builds? On the other hand, I confirm using this branch properly shows content in a console window.

If you launch debug build of current `main` (without this PR) from console (i.e. by running `npx tauri dev`) that's expected cause then that console window acts as your console. If you launch it from a release build it will show the empty console.

In contrast this branch should always spawn Northstar + Northstar console window, regardless of debug or release builds ^^

Alystrasz (approved):

PR #114, opened 2022-12-15
fix: Mod view with no game install

### LINKED ISSUE #64: Mod tab throws error when opening it without a selected Titanfall2 install
Mod tab throws error when opening it without a selected Titanfall2 install. Instead it should simply show something like "No install found" or "No mods detected" or something along the lines.

Alystrasz:
Display a message on empty mods view instead of showing an error notification.
Closes #64.

##### Before

Capture d’écran de 2022-12-15 08-25-10

##### After

Capture d’écran de 2022-12-15 08-23-40

### DIFF
--- a/src-vue/src/plugins/store.ts
+++ b/src-vue/src/plugins/store.ts
@@ -238,6 +238,13 @@ export const store = createStore<FlightCoreStore>({
                 game_path: state.game_path,
                 install_type: state.install_type
             } as GameInstall;
+
+            // If there's no game path, prevent looking for installed mods.
+            if (state.game_path === undefined) {
+                console.warn('Cannot load installed mods since so game path is selected.');
+                return;
+            }
+
             // Call back-end for installed mods
             await invoke("get_installed_mods_caller", { gameInstall: game_install })
                 .then((message) => {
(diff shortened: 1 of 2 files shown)


src-vue/src/plugins/store.ts
```diff
@@ -245,12 +245,16 @@ export const store = createStore<FlightCoreStore>({
...
+                    // TODO discuss if we should delete this
+                    // If there's no other error case as "no install selected", I think we can get rid of it.
+                    /*
                     ElNotification({
                         title: 'Error',
                         message: error,
                         type: 'error',
                         position: 'bottom-right'
                     });
+                    */
```
GeckoEidechse, line 257: While unlikely in practice, there can be other error cases. E.g.: lack of mods directory or some incorrectly formatted JSON.

It might make more sense to do an if-guard to check if the install path is actually set, before calling the backend function ^^

GeckoEidechse (commented):
Current implementation works on both Windows and Linux ^^

(though if-guard before calling backend would probably be preferred)

src-vue/src/plugins/store.ts
Alystrasz: Reverted in 2f6f90ee0dcafb03853b9dd29f79c5c118dcd350

GeckoEidechse (approved):
Changes look good and code was (re-)confirmed working in testing on Linux ^^

PR #115, opened 2022-12-16
feat: Alternatively check if EA App is running

GeckoEidechse:
With EA switching from Origin to EA App, we need to check for that instead.

Longterm, we should just launch via Steam and/or EA App directly, passing `-norhtstar` in the process to get it to launch Northstar instead.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -311,6 +311,12 @@ pub fn check_origin_running() -> bool {
     for _process in s.processes_by_name("Origin.exe") {
         // check here if this is your process
         // dbg!(process);
+        // There's at least one Origin process, so we can launch
+        return true;
+    }
+    // Alternatively, check for EA Desktop
+    for _process in s.processes_by_name("EADesktop.exe") {
+        // There's at least one EADesktop process, so we can launch
         return true;
     }
     false


GeckoEidechse:
7d9a78d53384728bd72af5572f07affddfe18ecb resolved not being able to run Northstar due to "Origin not running". They owned Titanfall2 on Steam but that nowadays also runs via EA App.

https://discord.com/channels/920776187884732556/1047303188413423637/1053487167504138300

GeckoEidechse:
Had this as draft cause I wanted to change function naming / error message still etc. But giving that it's tested to work and is blocking some people from playing, this should be merged now and naming fixed in separate PR.

Alystrasz:
Should I install EA app to test this?
(I heard that doing this would nuke my Origin install)

GeckoEidechse:
> Should I install EA app to test this?

It was tested to work already so no need to. Just put you as reviewer so that you're aware and can double check the code (though it's like 5 lines so there's not too much to check ^^)

&nbsp;

> (I heard that doing this would nuke my Origin install)

Exactly, it's a one way process so unless you're ready to switch I wouldn't recommend it :P

Alystrasz:
Gotta trust you on this one then!
I assume that FC will display "Origin is running: true" when EA app is opened, right?

GeckoEidechse:
> Gotta trust you on this one then! I assume that FC will display "Origin is running: true" when EA app is opened, right?

Yup, hence #116. But given that EA is already moving people to EA App, I want to get this out and fix the naming once I get to it ^^"

Alystrasz (approved):

PR #119, opened 2022-12-22
feat: Maximize button + clean menubar layout

Alystrasz:
This adds a new window control that enables user to maximize/minimize window.

It also restructures the menu bar, to properly display menu items when menu bar width reduces.

---

###### Maximizing (I couldn't capture entire screen, hence top screen cut, needed for me to be able to minimize window :) )

Peek 22-12-2022 16-43

###### Before

Peek 22-12-2022 16-35

###### After

Peek 22-12-2022 16-33


GeckoEidechse (approved):
Confirmed working in testing. Cannot say much about code itself but it works so it should be fine ig ¯\_(ツ)_/¯

GeckoEidechse:
(Also I absolutely love the GIF showcase <3)

PR #121, opened 2022-12-22
RFC: Variable tab bar demo

GeckoEidechse:
:warning: **This is a draft PR to showcase the potential of having variable number of entries in the tab bar.** :warning: 

## Idea:

By default only show base tabs but offer various buttons to open other views. Once opened this views will stay available on the tab bar until the end of the session (i.e. until the application is restarted).

## Advantages:

- Switching between views is instant. E.g. switching from Thunderstore mod install to launching game is just 2 clicks (switch to `PlayView`, click on _launch game_). This is quicker compared to e.g. Viper where browsing mods opens a pop-up overlay.
- No overwhelming new users by not showing all views "immediately"
- No need for any subpages/subviews. Just add another tab :P
- As usually not all tabs will be opened in one session, the tab bar will stay small in general.
- Anything else?

## Disadvantages:
- If buttons to open hidden views are not well placed, users might never discover them
- Variable number of tab entries bad?
  (I don't see the issue personally but I might just be missing sth ^^)
- Anything else?

Rendered:

https://user-images.githubusercontent.com/40122905/209241721-3d81ef59-8c5f-4ed8-9ce8-1c50d461548b.mp4

### **Important note:** The views to hide by default were chosen to be the ones that I deemed to be used less often. This does not mean we should actually hide these specific views by default on main. This was just for the sake of the demo. For example Thunderstore view is way too pretty to hide it :P


GeckoEidechse:
@Alystrasz thoughts? :>

This is basically in relation to #72 as I'm not too much of a fan of putting RepairView into some submenu ^^

Alystrasz:
Even if your demo works technically, this feature goes against basic UX principles.

1. **Menu should be fixed and not change during app lifetime**: you told it yourself, users might be lost if one day they see an item that isn't present the next day. Big gaming softwares like Steam or Battle.net don't change their menus over time, we shouldn't do it either. Only tolerable exception (to me) is the dev section, since it's not meant for normal users.
2. **FlightCore is not a web browser**: I know that its frontend is made using web technologies, but it's not meant to have several "tabs" like you would have on Chrome for instance.
3. **Menu space is not a problem**: I think we'll agree we don't want to see the "..." submenu appear; I plan to remove the Thunderstore section anyways and merge it into the mods section.

##### Additional comments

> E.g. switching from Thunderstore mod install to launching game is just 2 clicks (switch to PlayView, click on launch game)

I agree minimizing clicks count is important, but it's not linked with the "variable tabs" feature.

> This is quicker compared to e.g. Viper where browsing mods opens a pop-up overlay.

Agreed! The Thunderstore popup is even kinda hidden, I don't think all Viper users know they can install mods from the Internet.

> No need for any subpages/subviews. Just add another tab :P

You seem scared about subviews, UI frameworks have a lot of great components to handle those 😄

GeckoEidechse:
> You seem scared about subviews, UI frameworks have a lot of great components to handle those

It's less the technical aspect but the UI flow one. I'd prefer to switch between different UI elements in as few clicks as possible. I have waiting for any form of animation to finish just so I can click on the next submenu :P

My basic train of thought is basically I wanna be able to switch from one View to another with a single click while also not showing all views at once. The RepairView is the best example for this. It should be rarely needed so putting it on the main tab bar does not make sense. At the same time, after applying a "repair" you wanna check on it ASAP so RepairView -> PlayView should be a single click... ^^

I'm starting to think maybe we should switch to vertical tabs like VTOL...



> >  E.g. switching from Thunderstore mod install to launching game is just 2 clicks (switch to PlayView, click on launch game)

> I agree minimizing clicks count is important, but it's not linked with the "variable tabs" feature.

fair

> I plan to remove the Thunderstore section anyways and merge it into the mods section.

Not sure how that would work out but I'm open for design ideas.  I'd suggest only PRing a mvp as to not "lose" too much time in case it doesn't fit ^^

In particular opening mods tab should not load all mods from Thunderstore each time and it should be easy to tell which mods are installed. Not to mention that `nsmod != tsmod` ^^

GeckoEidechse:
Closing RFC as we went with integrating ThunderstoreView into ModsView and opening new window for repair instead.

PR #124, opened 2023-01-04
feat: Show newest version number

GeckoEidechse:
Show version number of new release in outdated warning.

demo screenshot


GeckoEidechse:
Adds a new TypeScript object in `src-vue/src/utils/FlightCoreVersion.d.ts` that mirrors the Rust object. As discussed, uh _somewhere_, we wanna auto-generate TypeScript objects from Rust code directly but I haven't looked into that yet and I wouldn't block the PR because of it ^^

Alystrasz (approved):
Confirmed working on Linux.
LGTM.

PR #125, opened 2023-01-04
fix: Add empty folder to stop macro complaining

GeckoEidechse:
A Tauri macro complains when the dist folder does not exist. Adding a `.gitkeep` means that cloning the repo, the folder will be created.

Talking about this macro error
image

Note that the macro error does not prevent compilation in any way. It's purely an annoyance thing :P


GeckoEidechse:
Actually adding that `.gitkeep` does "not work" in the sense that it gets deleted when a build is triggered which will show up in the diff and might get added when someone commits.
So in other words, someone making changes, building a release (`npx tauri build`), and then adding all changes will delete the `.gitkeep` again in their commit.

@Alystrasz any other ideas how to get the folder to exist when cloning (or at least on running `npx tauri dev`)?

Alystrasz:
Is this issue still a problem?
I can't reproduce it on my side.

image

GeckoEidechse:
> Is this issue still a problem?
> I can't reproduce it on my side.

Yes. It's cause once you run `npx tauri build` just once or any other action that creates `src-vue/dist`, the folder exists and it no longer complains. This means you won't notice it unless you do a fresh clone or run it in e.g. CI where the environment is always fresh. In fact, I [had to add special handling in CI](https://github.com/R2NorthstarTools/FlightCore/blob/49c454a0c5296c42229e84b2f1f23e336e4d8aea/.github/workflows/push-test.yml#L38-L39) for creating the folder first cause running `cargo test` also makes sure that macros do not panic ^^

GeckoEidechse:
Turns out it's not just the IDE complaining but actually gives an error on a complete fresh clone when only compiling Rust code first:

image

Alystrasz:
Looks like there's no need to create this directory anymore:

image

PR #127, opened 2023-01-04
refactor: Deserialize into struct directly

GeckoEidechse:
Instead of deserializing into a generic serde struct and then extracting specific value from it, deserialize directly into the struct we want.

Basically less lines of code thanks to serde :sparkles: magic :sparkles:


Alystrasz (approved):
Confirmed working on Windows.

PR #128, opened 2023-01-04
refactor: Remove unnecessary caller function pattern

GeckoEidechse:
Removes the unnecessary `function_caller` pattern by just adding the `#[tauri::command]` macro to the corresponding function directly.

I couldn't remove the pattern for all functions cause some seem to throw weird compiler macro errors when I do. Needs further investigating but the adjusted functions in this PR all work fine ^^


Alystrasz (approved):
A good code refactoring PR is always welcome :)

PR #129, opened 2023-01-04
feat: Allow opening a window with various repair features

GeckoEidechse:
This window contains various common repair tasks like disabling all but core mods, deleting FlightCore persistent store, etc.

Note that for this PR I purposefully only moved a single one of the currently existing repair buttons to the new view in order to keep the diff small.

The rest will be moved in a separate PR after this one has been merged.

The button to open the repair window will be moved from the DeveloperView to the settings page in a separate PR once the rest of the functions have been moved ^^

Rendered:
image


GeckoEidechse:
For some reason I cannot get the window to launch on a specific view >.<

GeckoEidechse:
> For some reason I cannot get the window to launch on a specific view >.<

Turns out cause of the router it needs to be prefixed with `/#`, e.g. `/#/changelog`

GeckoEidechse:
Okay I cannot get the top nav bar to not be visible when launching new window. Trying to hid it with CSS in `RepairView.vue` via

```html
<style scoped>
#fc_menu-bar {
    visibility: hidden !important;
}
</style>
```
and removing the `scoped` keyword just means it now affects all views and it's hidden everywhere...

GeckoEidechse:
This is how it looks currently

image


Naturally the menu bar shouldn't be in the new window that exists just for repair stuff.

src-vue/src/views/RepairView.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+    <div class="fc-container">
```
Alystrasz, line 2: This should have no padding-top, since there's no menu on this view.

src-vue/src/views/DeveloperView.vue
```diff
@@ -207,7 +186,26 @@ export default defineComponent({
             await persistentStore.clear();
             // ...and save
             await persistentStore.save();
-        }
+        },
+        async createNewWindow() {
+            await invoke("open_repair_window").then((message) => {
```
Alystrasz, line 191: I don't think we should show notifications here.

src-tauri/src/main.rs
```diff
@@ -345,3 +346,16 @@ async fn get_server_player_count() -> Result<(i32, usize), String> {
 
     Ok((total_player_count, server_count))
 }
+
+#[tauri::command]
+/// Spawns repair window
+async fn open_repair_window(handle: tauri::AppHandle) -> Result<(), String> {
+    let _repair_window = tauri::WindowBuilder::new(
```
Alystrasz, line 353: You should name the window something different than "tao window".

Alystrasz (changes requested):
Looking good!
After a few modifications, I'll be glad to merge this.

src-tauri/src/main.rs
GeckoEidechse: Done in ce8717f314e8c4f85169f712e75d99b97845dc65

It's now called `FlightCore Repair Window`

src-vue/src/views/DeveloperView.vue
GeckoEidechse: Removed in 50bf106284f28211b4f268153c8201f467104d34

src-vue/src/views/RepairView.vue
GeckoEidechse: Fixed in e7c6746a17a8ec0d9fa770be24cf0c0940330343

Alystrasz (commented):
Currently, repair window does not close when closing main window; I think it should.
What are your thoughts about this?

GeckoEidechse:
> Currently, repair window does not close when closing main window; I think it should. What are your thoughts about this?

Agreed. I didn't account for that scenario. Will investigate ^^

GeckoEidechse:
Closing all windows is now implemented in 20f1db85b838ad8cd4f920b571245bda2850246c

Alystrasz (approved):
Code looks good and works as described.
Confirmed running/building on Windows.

PR #130, opened 2023-01-05
feat: Display Northstar playercount and servercount in frontend

GeckoEidechse:
Just playercount and server count on PlayView in an unobtrusive way.

Depends on

- [x] #131
- [x] #135

TODO:
- [x] Figure out placement location

Rendered:
image


src-vue/src/views/PlayView.vue
```diff
@@ -32,6 +39,11 @@ export default defineComponent({
             <div v-if="northstarVersion !== ''" class="fc_changelog__link" @click="showChangelogPage">
                 (see patch notes)
             </div>
+            <br />
+            <div>
+                {{ playerCount }} Players,
+                {{ serverCount }} Servers
```
GeckoEidechse, line 45: This does not handle the singular case but honestly I don't think it's worth handling as the `1 player` or `1 server` case is super rare. Looking at player numbers it would only happen when Northstar just got version gated and people are updating. So far we had 3 version gates and the first few players always updated within seconds. ^^

src-vue/src/plugins/store.ts
```diff
@@ -65,6 +68,9 @@ export const store = createStore<FlightCoreStore>({
             northstar_is_running: false,
             origin_is_running: false,
 
+            player_count: -1,
+            server_count: -1,
```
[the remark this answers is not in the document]
Alystrasz: Nah, it should still be a number; maybe 0?

src-vue/src/views/PlayView.vue
Alystrasz: I think it's ok not to handle this case as well.

src-vue/src/plugins/store.ts
```diff
@@ -382,6 +388,16 @@ async function _initializeApp(state: any) {
         // Check installed Northstar version if found
         await _get_northstar_version_number(state);
     }
+
+    await invoke("get_server_player_count")
+        .then((message) => {
+            state.player_count = (message as any)[0];
+            state.server_count = (message as any)[1];
+        })
```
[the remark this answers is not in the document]
Alystrasz: Here's the declaration of the `invoke` method:
```typescript
declare function invoke<T>(cmd: string, args?: InvokeArgs): Promise<T>;
```

The way you're calling it right now (`invoke("get_server_player_count")`) doesn't set the `T` type, so returned promise will itself return an `unknown`, which prevents you to use return value as you would like to.

To fix this, since you know this `invoke` call will return an array of two number, you can summon it with according type:
```typescript
await invoke<[number, number]>("get_server_player_count")
        .then((message) => {
            state.player_count = message[0];
            state.server_count = message[1];
        });
```

src-vue/src/plugins/store.ts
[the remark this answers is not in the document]
GeckoEidechse: In that case I would just stick with `-1` tbh. Cause it indicates more clearly that something went wrong compared to `0` which is valid player/server count. Sure we're unlikely to ever hit that (hopefully) but if some API error causes FlightCore to be unable to load the playercout, the user will see `0 players, 0 servers` and might just think that Northstar is _"dead"_.

src-vue/src/plugins/store.ts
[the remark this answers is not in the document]
GeckoEidechse: Thanks for the detailed explanation <3

Done in a8b0f44747afcdd880f6d1ce2d8a2b350ead4311

src-vue/src/plugins/store.ts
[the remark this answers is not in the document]
Alystrasz: I don't think users should notice when something goes wrong anyway; FC displaying `-1` will mean to both of us that something went wrong while fetching stats, users will rather think that FC is broken :/
Whatever value you choose, please display a message instead of raw value.

src-vue/src/views/PlayView.vue
```diff
@@ -32,6 +39,11 @@ export default defineComponent({
             <div v-if="northstarVersion !== ''" class="fc_changelog__link" @click="showChangelogPage">
                 (see patch notes)
             </div>
+            <br />
+            <div>
```
Alystrasz, line 43: Unsure about positioning of these tbh, left-hand part of this view relates to local Northstar install (version + launch button), stats feel a bit out of place.

src-vue/src/views/PlayView.vue
```diff
@@ -32,6 +39,11 @@ export default defineComponent({
             <div v-if="northstarVersion !== ''" class="fc_changelog__link" @click="showChangelogPage">
                 (see patch notes)
             </div>
+            <br />
```
Alystrasz, line 42: You should apply spacing with CSS rather than using this ugly tag :)

Alystrasz (commented):
Also, would it be greater if those statistics refreshed in pseudo real-time?

src-vue/src/plugins/store.ts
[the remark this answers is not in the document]
GeckoEidechse: Done in f2d4952c9fc98d8bc547d1ade7136ad621207410

Rendered:
image

src-vue/src/views/PlayView.vue
GeckoEidechse: Kinda agree but also didn't know where else to put it ^^"

src-vue/src/views/PlayView.vue
GeckoEidechse: My CSS is a bit rusty, do you maybe wanna do that part? ^^"

Alystrasz: Yay! Gotta find a good way to display stats first though.

Alystrasz: @GeckoEidechse still need help for adding margin to this? :)

src-vue/src/views/PlayView.vue
Alystrasz: I really have no other idea for where to put it, so let's put it there.

src-vue/src/views/PlayView.vue
GeckoEidechse: Uh, yes please ^^"

src-vue/src/views/PlayView.vue
```diff
@@ -69,6 +82,10 @@ export default defineComponent({
     color: rgb(168, 168, 168);
 }
 
+.fc-stats__container {
+    margin-top: 3px;
```
Alystrasz, line 86: @GeckoEidechse do you think this is enough spacing?

src-vue/src/views/PlayView.vue
```diff
@@ -32,6 +38,13 @@ export default defineComponent({
             <div v-if="northstarVersion !== ''" class="fc_changelog__link" @click="showChangelogPage">
                 (see patch notes)
             </div>
+            <div v-if="playerCount >= 0" class="fc-stats__container">
+                {{ playerCount }} Players,
```
Alystrasz, line 42: I think we should not use uppercase here

src-vue/src/views/PlayView.vue
```diff
@@ -32,6 +38,13 @@ export default defineComponent({
...
+                {{ serverCount }} Servers
```
Alystrasz, line 43: I think we should not use uppercase here neither

Alystrasz (dismissed):

src-vue/src/views/PlayView.vue
GeckoEidechse: Looks fine to me ^^

image

src-vue/src/views/PlayView.vue
GeckoEidechse: Changed to lowercase in ea1efa939681e79e4b2ea513bf683628adc63d14

src-vue/src/views/PlayView.vue
GeckoEidechse: Changed to lowercase in ea1efa939681e79e4b2ea513bf683628adc63d14

GeckoEidechse (commented):
Looks good from my side now

(cannot approve my own PR even if it contains someone else's changes)

PR #131, opened 2023-01-05
refactor: Store global const in separate file

GeckoEidechse:
Moved user agent there for now.

Storing all global consts in a single file makes adjusting them easier as one also have to look at a single place. This is especially useful for anyone wanting to make their own spin/fork of the application (NorthstarCN _\*hint\*_ _\*hint\*_ ;P)


Alystrasz (approved):
LGTM.
Do you plan to add new constants to this new module? If not, you can merge this.

GeckoEidechse:
Not at the moment but future PRs might of course make use of this ^^

PR #133, opened 2023-01-12
refactor: Bump libthermite

GeckoEidechse:
This updates to version `0.4.0` of [`libthermite`](https://crates.io/crates/libthermite).

In the process, we now also no longer write the Thunderstore mod string to the Northstar mod's `mod.json` but rather copy the `manifest.json` from the Thunderstore mod into the Northstar mod and add a file called `thunderstore_author.txt` that contains the author name of the Thunderstore mod to uniquely map a Northstar mod to a Thunderstore mod.
This is done to increase compatibility with Viper and VTOL. See also the discussion here: https://github.com/0neGal/viper/issues/165

Reading `mod.json` for Thunderstore string is still supported for the time being. In the future it will be changed to a "converter" and then later will be fully deprecated.

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -331,8 +331,11 @@ pub async fn fc_download_mod_and_install(
         Err(e) => return Err(e.to_string()),
     };
 
+    // Get Thunderstore mod author
+    let author = thunderstore_mod_string.split("-").next().unwrap();
+
     // Extract the mod to the mods directory
-    match thermite::core::manage::install_mod(&f, std::path::Path::new(&mods_directory))
+    match thermite::core::manage::install_mod(author, &f, std::path::Path::new(&mods_directory))
     {
         Ok(()) => (),
         Err(err) => return Err(err.to_string()),
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -237,32 +237,6 @@ async fn get_ns_mod_download_url(thunderstore_mod_string: String) -> Result<Stri
     Err("Could not find mod on Thunderstore".to_string())
 }
 
-/// Adds given Thunderstore mod string to the given `mod.json`
-/// This way we can later check whether a mod is outdated based on the TS mod string
-fn add_thunderstore_mod_string(
-    path_to_mod_json: String,
-    thunderstore_mod_string: String,
-) -> Result<(), anyhow::Error> {
-    // Read file into string and parse it
-    let data = std::fs::read_to_string(path_to_mod_json.clone())?;
-    let parsed_json: serde_json::Value = json5::from_str(&data)?;
-
-    // Insert the Thunderstore mod string
-    let mut parsed_json = parsed_json.as_object().unwrap().clone();
-    parsed_json.insert(
-        "ThunderstoreModString".to_string(),
-        serde_json::Value::String(thunderstore_mod_string),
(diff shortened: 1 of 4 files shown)


src-tauri/src/mod_management/mod.rs
```diff
@@ -133,6 +141,7 @@ fn parse_mod_json_for_mod_name(mod_json_path: String) -> Result<String, anyhow::
 }
 
 /// Parses `mod.json` for Thunderstore mod string
+/// This is a legacy function as nowadays `manifest.json` should be in Northtar mods folder and read from there
```
Alystrasz, line 144: Maybe we can even more explicitely set this as deprecated?
https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

src-tauri/src/mod_management/mod.rs
```diff
@@ -153,6 +162,33 @@ fn parse_mod_json_for_thunderstore_mod_string(
...
+    let data = std::fs::read_to_string(manifest_json_path)?;
+    let thunderstore_manifest: ThunderstoreManifest = json5::from_str(&data)?;
+
+    // Check if `thunderstore_author.txt` exists and parse
+    let mut file = std::fs::File::open(ts_author_txt_path)?;
+    let mut thunderstore_author = String::new();
+    file.read_to_string(&mut thunderstore_author)?;
+
+    // Build mod string
+    let thunderstore_mod_string = format!("{}-{}-{}", thunderstore_author, thunderstore_manifest.name, thunderstore_manifest.version_number);
```
Alystrasz, line 187: Does this handle names with spaces? (IIRC, Thunderstore replaces them with underscores)

GeckoEidechse: Yesn't we already write it down padded with underscores in the name ^^

So the `thunderstore_author.txt` will always have spaces replaced with underscores.

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Hmm, this is more designed for libraries and will throw a compiler warning on every build as long as the function is still in use.

Not super a fan of clogging build with a warning each compile, so as long as the function is still supported, which it is right now, I wouldn't add a deprecation macro to it ^^

GeckoEidechse: Hmm I was thinking maybe setting it to a version number greater then current to indicate deprecation in the future but that still causes the warning to show up ^^"

Alystrasz: Doesn't this show a warning if the deprecated method is used instead?

GeckoEidechse: It shows a warning on compile if the deprecated function is used somewhere else in the code. Given that we still need to use it to map Northstar mods to Thunderstore mod that were installed using the old method we will have to call it. So that function technically isn't really deprecated yet. It just exists for legacy compatibility.

The moment the user reinstalls or updates a mod installed using the old method (i.e. writing thunderstore mod string to `mod.json`) it will be overwritten to use the new method (adding `manifest.json` and `thunderstore_author.txt`).

The logic that writes to `mods.json` is not just deprecated but completely removed with this PR as we no longer need it ^^

Alystrasz: Fine by me :)

Alystrasz (approved):
Confirmed working on Windows.

PR #134, opened 2023-01-12
refactor: Mods view

Alystrasz:
* Regroup local mods and Thunderstore mod views together
* Add search filters
  * Text filtering
  * Sorting
  * Mod categories

https://user-images.githubusercontent.com/11993538/212176934-1e99add5-0eea-4b0e-9db7-4bdae5403492.mp4


src-vue/src/utils/SortOptions.d.ts
```diff
@@ -0,0 +1,8 @@
+export enum SortOptions {
+    NAME_ASC = 'By name (A to Z)',
+    NAME_DESC = 'By name (Z to A)',
+    DATE_ASC = 'By date (from oldest)',
+    DATE_DESC = 'By date (from newest)',
+    MOST_DOWNLOADED = "Most downloaded",
+    TOP_RATED = "Top rated"
+}
```
GeckoEidechse, line 8: Missing newline >:(

src-vue/src/views/ThunderstoreModsView.vue
```diff
@@ -1,19 +1,11 @@
 <template>
-    <div class="fc-container">
+    <div class="fc-container" style="padding: 0">
```
GeckoEidechse, line 2: I assume the reason the padding is set to zero here instead of for `fc-container` is that this should only apply to this specific div?

src-vue/src/views/mods/LocalModsView.vue
```diff
@@ -0,0 +1,122 @@
+<template>
```
GeckoEidechse, line 1: Is there a reason this source file is in a different folder? ^^

`views/` vs `mods/`

src-vue/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "esModuleInterop": true,
-    "lib": ["ESNext", "DOM"],
+    "lib": ["ESNext", "DOM", "ES2018"],
```
GeckoEidechse, line 13: This forces _[ECMAScript 2018](https://learn.microsoft.com/en-us/openspecs/ie_standards/ms-es2018/38f64325-9d15-44c6-a5de-ec61a723b03f)_ right? Just so that I'm aware, why do we need to set this? ^^

GeckoEidechse (commented):
Looks amazing. Played around with it a bit and seems to work just fine. Noice stuff :D

A few comments below ^^

src-vue/tsconfig.json
Alystrasz: Visual Studio Code was showing promises' `finally` clauses as erroneous, unless I add this option.

src-vue/src/views/ThunderstoreModsView.vue
Alystrasz: Exactly

src-vue/src/utils/SortOptions.d.ts
Alystrasz: Addressed in c7eba636f434157024c6a69638a0a26b51c4aad4

src-vue/src/views/mods/LocalModsView.vue
Alystrasz: Oops, I forgot to put TS view into this folder as well 😄 
Addressed in 9fd76811f9c938fc5e12f4bc9162ad7890eef90f

src-vue/tsconfig.json
GeckoEidechse: Huh, I mean if it works then ¯\\\_(ツ)\_/¯

src-vue/src/plugins/modules/search.ts
```diff
@@ -0,0 +1,18 @@
...
+        // Selected mod categories
+        selectedCategories: [],
+        sortValue: {label: '', value: ''}
+    }),
+    getters: {
+        searchWords(state: SearchStoreState): string {
+            return state.searchValue.toLowerCase();
+        }
+    }
+  }
```
GeckoEidechse, line 18:
```suggestion
  }
  
```

src-vue/src/views/mods/LocalModsView.vue
```diff
@@ -0,0 +1,122 @@
...
+    },
+    mounted() {
+        this.$store.commit('loadInstalledMods');
+    }
+})
+</script>
+
+<style scoped>
+
+</style>
```
GeckoEidechse, line 122:
```suggestion
</style>

```

GeckoEidechse (commented):
Aight, took another look and it seems apart from two missing newlines, I already caught everything in the first review pass ^^

GeckoEidechse (approved):

PR #135, opened 2023-01-13
feat: Add backend code for getting playercount

GeckoEidechse:
This adds the backendcode for getting playercount as well as the necessary frontend code to call it on launch but unlike #130 does not actually show the player and server count as we are unsure where to display it.

This allows for fast tracking the backend code and such gets as to next release earlier which should also be the one that starts getting publicly advertised :eyes:

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -310,11 +310,6 @@ async fn clean_up_download_folder_caller(
     }
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone)]
-struct NorthstarServer {
-    #[serde(rename = "playerCount")]
-    player_count: i32,
-}
 
 /// Gets server and playercount from master server API
 #[tauri::command]
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -9,7 +9,7 @@ use std::{
     time::Duration,
 };
 
-use app::{*, constants::APP_USER_AGENT};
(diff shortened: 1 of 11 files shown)


GeckoEidechse:
Build failures right now are only Ubuntu and seem to be happening due to some repo being offline. To fix, just retrigger but it probably makes sense to wait a bit to ensure they will be online again. ^^

src-vue/src/plugins/store.ts
```diff
@@ -382,6 +388,16 @@ async function _initializeApp(state: any) {
         // Check installed Northstar version if found
         await _get_northstar_version_number(state);
     }
+
+    await invoke<[number, number]>("get_server_player_count")
```
Alystrasz, line 392: Didn't we speak about pinging master server every 5 minutes?

GeckoEidechse: Eh, I'll do it later, wanna get that release out :P

src-tauri/src/main.rs
```diff
@@ -307,3 +309,33 @@ async fn clean_up_download_folder_caller(
         Err(err) => Err(err.to_string()),
     }
 }
+
+
+/// Gets server and playercount from master server API
+#[tauri::command]
+async fn get_server_player_count() -> Result<(i32, usize), String> {
+
+    let server_browser_endpoint = "/client/servers";
```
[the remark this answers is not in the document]
GeckoEidechse: Good point, done in 5f51f3cafcd2e746218fd30acec394ba09c9b9be

PR #139, opened 2023-01-19
feat: Allow installing PRs from DeveloperView

GeckoEidechse:
Allows installing launcher and mod PRs from within FlightCore, essentially replacing https://github.com/GeckoEidechse/northstar_dev_testing_helper_tool/

TODO

- [x] Code cleanup
   - [x] General stuff
   - [x] Remove duplicate code
   - [x] Remove unnecessary API calls
   - [x] Move consts to `constants.rs`
- [x] Add clickable link to PR
- [x] Better UI (thanks @Alystrasz <3)
- [x] ~~Better error messages when PR couldn't be downloaded (e.g. https://github.com/R2Northstar/NorthstarLauncher/pull/414)~~ Solved by fixing underlying issue in c4fc200af92fff7a8a2594810fe18d84f01be772
- [x] Testing
  - [x] Launcher install
  - [x] Mods install

**Requires:**
- [x] #140 
- [x] #150


GeckoEidechse:
Ok this is nearing completion now. Only thing missing is reducing duplicate code, around launcher vs mod, i.e. use an enum to discern the two and then have a single function to handle both.

Alystrasz (commented):
Before reviewing in details, I have a few questions:
* Why can't I install some PRs? ("Couldn't grab download link for PR 414" isn't an explicit error)
* Can we prevent installing several times the same PR?
* While installing mods PRs, instead of forcing users to go outside FlightCore to switch profile and actually test them, we should handle the profile switch ourselves, shall we?
* Do you want me to improve UI a bit?

GeckoEidechse:
> Before reviewing in details, I have a few questions:
> 
> * Why can't I install some PRs? ("Couldn't grab download link for PR 414" isn't an explicit error)

I guess no downloadable CI build. Need to investigate further

> * Can we prevent installing several times the same PR?

We could but that would require tracking which PR is installed right now which would be a lot more additional logic that (in my opinion) would add rather little value. At least to me there's no point in preventing installing the same PR multiple times. ^^

> * While installing mods PRs, instead of forcing users to go outside FlightCore to switch profile and actually test them, we should handle the profile switch ourselves, shall we?

Would require fully supporting profiles first, which we don't do yet. Next best solution would be having a button to launch `NorthstarLauncher.exe -profile=R2Northstar-PR-test-managed-folder` but that would require rewriting some Northstar launching code which kinda felt outside of the scope of this PR ^^

Overall, same with the previous question, the goal of this PR is to achieve feature parity with https://github.com/GeckoEidechse/northstar_dev_testing_helper_tool so that I can deprecate that. The PR in it's current state has already achieved that minus search filter, which leads me to:

> * Do you want me to improve UI a bit?

If you don't mind <3

GeckoEidechse:
> Do you want me to improve UI a bit?

For now I added a notification similar to #154 where a notification is displayed once installation starts and removed again upon completion and replaced with one that says installation has completed.

With that out of the way, I'd say this PR is ready to merge and I'd turn the last TODO into an issue to solve post-merge.

The reason for this being that the branch is starting to diverge more and more and it's already become annoying to resolve merge conflicts every time another PR gets merged.

Additionally this feature is already on par with https://github.com/GeckoEidechse/northstar_dev_testing_helper_tool/ and it's too useful to keep delaying it for polish IMO ^^

src-vue/src/components/PullRequestsSelector.vue
```diff
@@ -0,0 +1,186 @@
...
+        onChange(e: Object) {
+            const openedCollapseNames = Object.values(e);
+            if (openedCollapseNames.includes('1') && this.pull_requests_launcher.length === 0) {
+                this.getPullRequests('LAUNCHER');
+            }
+            if (openedCollapseNames.includes('2') && this.pull_requests_mods.length === 0) {
+                this.getPullRequests('MODS');
+            }
+        },
+        async getPullRequests(pull_request_type: PullRequestType) {
```
Alystrasz, line 80: I think that for more clarity, we should create a store module dedicated to PR installation with this method.

src-vue/src/components/PullRequestsSelector.vue
```diff
@@ -0,0 +1,186 @@
...
+                .catch((error) => {
+                    ElNotification({
+                        title: 'Error',
+                        message: error,
+                        type: 'error',
+                        position: 'bottom-right'
+                    });
+                });
+        },
+        async installLauncherPR(pull_request: PullsApiResponseElement) {
```
Alystrasz, line 105: I think that for more clarity, we should create a store module dedicated to PR installation with this method.

src-vue/src/components/PullRequestsSelector.vue
```diff
@@ -0,0 +1,186 @@
...
+                        type: 'error',
+                        position: 'bottom-right'
+                    });
+                })
+                .finally(() => {
+                    // Clear old notification
+                    notification.close();
+                });
+        },
+        async installModsPR(pull_request: PullsApiResponseElement) {
```
Alystrasz, line 139: I think that for more clarity, we should create a store module dedicated to PR installation with this method.

Alystrasz (changes requested):
We should create a store module dedicated to PR installation (just like `searchModule` module) to keep code clean.
Otherwise, I confirm code looks great and c4fc200af92fff7a8a2594810fe18d84f01be772 fixed related issue; I'll be happy to merge once comments have been addressed :)

src-vue/src/components/PullRequestsSelector.vue
GeckoEidechse: So similar to the [`searchModule`](https://github.com/R2NorthstarTools/FlightCore/blob/78a9522e9ddae501b4ea1b4cd094a9a959b029f2/src-vue/src/plugins/modules/search.ts)?

GeckoEidechse: Oh nevermind you already wrote that above xD

GeckoEidechse: Ok, I moved to the store location to the submodule in 7d2b65b039a7bade3ce8478d03cc6b20846d6e34 but I'm not sure how to do move the methods as they are not `getters` (?) and not sure what property to use then... ^^"

Alystrasz: You can move methods there as `mutations`, and then summon them using `commit` calls.
Documentation here: https://vuex.vuejs.org/guide/mutations.html

GeckoEidechse: For moving `installLauncherPR` and `installModsPR` I'm not quite sure how to access `game_path` from the parent module store.

Alystrasz: You can import the `store` object and access its state:
```javascript
import {store} from "../store";
console.log(store.state.game_path);
```

src-vue/src/components/PullRequestsSelector.vue
GeckoEidechse: Done in 7ca08697a75f7f7adf791b7baba09c059bd8c392

src-vue/src/components/PullRequestsSelector.vue
GeckoEidechse: Done in 068d1e56ab428b4b94bfeab9943019a4f26db9e8

src-vue/src/components/PullRequestsSelector.vue
GeckoEidechse: Done in 92d0d2009887a67353b2cee71b6c673fc6e5d7ca and 75f7deed46a9bc69da297060afed2cdc2e162737

Alystrasz (approved):
Comments were addressed, code looks good and works well.
LGTM.

PR #141, opened 2023-01-20
fix: Properly parse mod string

GeckoEidechse:
Instead of doing a simply string replace, we now parse the mod string into an object.


Alystrasz (approved):
Confirmed working on Windows.

PR #142, opened 2023-01-20
fix: Remove unused serde import in `main.rs`

GeckoEidechse:
Seems like I forgot to remove it when I did some code refactoring. Removes a compiler warning about unused imports.


Alystrasz (approved):
LGTM.

PR #143, opened 2023-01-20
chore: Autoformat Rust source files

GeckoEidechse:
Autoformatted all Rust source code that wasn't formatted properly.

Maybe we should start enforcing formatting via CI? ^^


ASpoonPlaysGames:
If you want to enforce formatting in any way always automate it with CI

Alystrasz (approved):

PR #146, opened 2023-01-23
feat: Add Rust code format check to CI

### LINKED ISSUE #144: Enforce code formatting via CI
In order to avoid any unformatted PRs getting merged.

GeckoEidechse:
Performs source code formatting check in CI and fails if the formatting is wrong.

TODO:

- [x] change it so that it runs in parallel next to other tasks.

Closes #144

### DIFF
--- a/.github/workflows/push-test.yml
+++ b/.github/workflows/push-test.yml
@@ -51,9 +51,8 @@ jobs:
     - name: format-check
       uses: actions-rs/cargo@v1
       with:
-        path: src-tauri
         command: fmt
-        args: --all -- --check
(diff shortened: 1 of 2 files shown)


.github/workflows/push-test.yml
```diff
@@ -14,6 +14,22 @@ jobs:
         run: |
           python3 scripts/check_version_numbers.py
 
+  # Ensure correct Rust code formatting
+  fmt:
```
Alystrasz, line 18:
```suggestion
  formatting:
```
Just to be sure everybody understands what this action does :)

PR #149, opened 2023-01-28
fix: Set current route as menu active item

### LINKED ISSUE #26: Visual bug with router refactor
Seems to work fine in testing for the most part. `mounted()` no only gets called when the view is actually opened, i.e. what we wanted to achieve with this PR ^^

One issue I noticed is that when enabling/disabling dev mode, the highlighted tab at the top gets updated incorrectly.
In particular, enabling dev mode highlights the _DEV_ tab even though it isn't the open one:
screenshot

and disabling dev mode switches back to _PLAY_ view but doesn't highlight it:
screenshot

_Originally posted by @GeckoEidechse in https://github.com/GeckoEidechse/FlightCore/pull/25#pullrequestreview-1148934672_

Alystrasz:
Closes #26.

### DIFF
--- a/src-vue/src/App.vue
+++ b/src-vue/src/App.vue
@@ -50,7 +50,7 @@ export default {
     <nav id="fc_menu-bar">
       <!-- Navigation items -->
       <el-menu
-        default-active="/"
+        :default-active="$route.path"
         router
         mode="horizontal"
         id="fc__menu_items"


GeckoEidechse (approved):
Confirmed working in testing on Windows.

Previously clicking on `see patch notes` in the main view would switch view but not update highlighted page in tab bar. With this PR it updates properly.

PR #150, opened 2023-01-30
refactor: Remove unnecessary use of anyhow

GeckoEidechse:
Single line change. No need for custom error type if we just return a string anyway.


GeckoEidechse:
Whoops, just noticed I "assigned" you instead of "requesting review", @Alystrasz 

Sorry about that ^^"

Alystrasz (approved):
Confirmed working on Windows.

PR #152, opened 2023-02-01
fix: Prevent mods view from crashing if Northstar is not installed

Jan200101:
`parse_installed_mods` will panic if Northstart mods folder doesn't exist so we check for that before trying to read from it.

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -196,7 +196,10 @@ fn parse_for_thunderstore_mod_string(nsmod_path: String) -> Result<String, anyho
 fn parse_installed_mods(game_install: GameInstall) -> Result<Vec<NorthstarMod>, String> {
     let ns_mods_folder = format!("{}/R2Northstar/mods/", game_install.game_path);
 
-    let paths = std::fs::read_dir(ns_mods_folder).unwrap();
+    let paths = match std::fs::read_dir(ns_mods_folder) {
+        Ok(paths) => paths,
+        Err(_err) => return Err("No mods folder found".to_string()),
+    };
 
     let mut directories: Vec<PathBuf> = Vec::new();
     let mut mods: Vec<NorthstarMod> = Vec::new();


src-tauri/src/mod_management/mod.rs
```diff
@@ -196,6 +196,11 @@ fn parse_for_thunderstore_mod_string(nsmod_path: String) -> Result<String, anyho
 fn parse_installed_mods(game_install: GameInstall) -> Result<Vec<NorthstarMod>, String> {
     let ns_mods_folder = format!("{}/R2Northstar/mods/", game_install.game_path);
 
+    // Check if mods folder exists
+    if !std::path::Path::new(&ns_mods_folder).exists() {
+        return Err("No mods folder found".to_string());
+    }
+
     let paths = std::fs::read_dir(ns_mods_folder).unwrap();
```
GeckoEidechse, line 204: Good catch. Was also able to confirm the issue on my end as well ^^

AFAIK in Rust doing error handling with `match` is preferred over separate check and then `unwrap`. So maybe something like this?

```suggestion
    let paths = match std::fs::read_dir(ns_mods_folder) {
        Ok(paths) => paths,
        Err(_err) => return Err("No mods folder found".to_string()),
    };
```

Jan200101: sure, I took this pattern from `get_log_list`

https://github.com/R2NorthstarTools/FlightCore/blob/fe0ada0c733687050a6a16d19d45a3de7790de7d/src-tauri/src/repair_and_verify/mod.rs#L71-L77

should I also replace that there?

GeckoEidechse: Yeah might be a shout, wrote that pretty early when my knowledge of common patterns in Rust was rather shallow ^^

Make it a separate PR though to keep things separated ^^

GeckoEidechse (approved):
Confirmed working in testing and code looks fine as well.

PR #155, opened 2023-02-02
refactor: Deal with `fs::read_dir` failure instead of checking if exists

Jan200101:
As discussed in #152

### DIFF
--- a/src-tauri/src/repair_and_verify/mod.rs
+++ b/src-tauri/src/repair_and_verify/mod.rs
@@ -71,6 +71,7 @@ pub fn clean_up_download_folder(
 pub fn get_log_list(game_install: GameInstall) -> Result<Vec<std::path::PathBuf>, String> {
     let ns_log_folder = format!("{}/R2Northstar/logs", game_install.game_path);
 
+    // List files in logs folder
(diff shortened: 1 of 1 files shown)


src-tauri/src/repair_and_verify/mod.rs
```diff
@@ -71,13 +71,10 @@ pub fn clean_up_download_folder(
...
     let ns_log_folder = format!("{}/R2Northstar/logs", game_install.game_path);
 
-    // Check if logs folder exists
-    if !std::path::Path::new(&ns_log_folder).exists() {
-        return Err("No logs folder found".to_string());
-    }
-
-    // List files in logs folder
-    let paths = std::fs::read_dir(ns_log_folder).unwrap();
+    let paths = match std::fs::read_dir(ns_log_folder) {
```
GeckoEidechse, line 74:
```suggestion
    // List files in logs folder
    let paths = match std::fs::read_dir(ns_log_folder) {
```

Restoring comment :P

GeckoEidechse (approved):
Aight, looks good to me now, merging when CI passes ^^

( I wish GH had a auto-merge when CI passes button like GitLab has >.< )

PR #158, opened 2023-02-05
refactor: Parse using serde deserialisation

GeckoEidechse:
This gets rid of a bunch of code by making use of [`serde`](https://crates.io/crates/serde) functionality. This should have honestly been written like this in the first place but when I first used `serde` I wasn't aware of that functionality even though it's its defining feature ^^"


Alystrasz (approved):
Good refactoring work!
LGTM, working on Linux.

PR #159, opened 2023-02-05
feat: Show mod version in LocalMods View

GeckoEidechse:
Added based on user feedback and tbh it also makes sense to show it ^^

Rendered:
image

Requires:
- [x] #158


GeckoEidechse:
PR is technically done but leaving as draft until #158 is merged cause it'll simplify the diff a lot for review ^^

GeckoEidechse:
CI is failing cause Ubuntu Azure repos are dead (again)

src-vue/src/views/mods/LocalModsView.vue
```diff
@@ -14,6 +14,7 @@
                     </template>
                 </el-popconfirm>
                 {{ mod.name }}
+                (v{{ mod.version }})
```
Alystrasz, line 17: Given you've defined version type as `Option<String>`, what happens here if `mod.version` is null?

GeckoEidechse: Just turns into nothing it seems.
image

Given that it is such a rare occasion should we add edge case handling (I think Northstar prints `0.0.0` as the version number then) or just ignore?

Alystrasz: If this happens, you shouldn't display version indicator at all.

GeckoEidechse: Aight, done with 5274aa9ada22d9299e87fb41c4d42214517a97fd

Alystrasz (approved):
LGTM, confirmed working on Linux.

PR #160, opened 2023-02-05
feat: Show TS icon for TS mods

GeckoEidechse:
In LocalModsView, show the Thunderstore icon for Northstar mods that have been installed from a Thunderstore mod.

Helps with seeing which mods come from Thunderstore.

Rendered:
image


src-vue/src/views/mods/LocalModsView.vue
```diff
@@ -14,6 +14,12 @@
                     </template>
                 </el-popconfirm>
                 {{ mod.name }}
+                <img
+                    v-if="mod.thunderstore_mod_string != null"
+                    src="/src/assets/thunderstore-icon.png"
+                    class="image"
+                    height="16"
```
GeckoEidechse, line 21: Probably not the right way to define size so I'm happy to hear what I should actually use ^^

Alystrasz: That's a good way to do it :)

Alystrasz (approved):
Confirmed working on Windows.

Issue #120, opened 2022-12-22
Expose core mods from backend to frontend


GeckoEidechse:
> Hmm, for that we need to signal core mods from backend to front-end. I don't think we have something for that yet... :/

_Originally posted by @GeckoEidechse in https://github.com/R2NorthstarTools/FlightCore/pull/110#discussion_r1054898148_

Alystrasz:
Without a need for signals, this is more a problem of code sharing between frontend and backend.
I think we should explore automatically building Typescript types, using a tool like https://github.com/Aleph-Alpha/ts-rs

GeckoEidechse:
> I think we should explore automatically building Typescript types, using a tool like https://github.com/Aleph-Alpha/ts-rs

Once #140 is merged which uses `ts-rs`, we could create a global const vector and expose that via `ts-rs` I think. Alternatively if that doesn't work the fall back would be to have a function call that returns the list of core mods.

GeckoEidechse:
Turns out you cannot use `ts-rs` for global consts. It's only for structs (which makes sense I guess). I think it'll make most sense to a have a call into backend that fetches the list of consts where necessary.

GeckoEidechse:
Actually, the only place where we declare some const a second time is in https://github.com/R2NorthstarTools/FlightCore/blob/a7e28562bb64e800f30b4f82a942ece2284b4121/src-vue/src/plugins/store.ts#L249 which is planned to be replaced with a call to backend anyway... ^^

PR #162, opened 2023-02-05
refactor: Move various consts to `constants.rs`

GeckoEidechse:
Move
- list of core Northstar mods
- list of blacklisted mods
- list of Titanfall2 Origin game IDs

to `constants.rs`

This should cover all remaining constants.


src-tauri/src/mod_management/mod.rs
```diff
@@ -12,8 +13,6 @@ use app::GameInstall;
 
 use json5;
 
-use crate::northstar::CORE_MODS;
-
 pub const BLACKLISTED_MODS: [&str; 3] = [
```
Alystrasz, line 16: Would it make sense to move those to constants as well?

GeckoEidechse: Done with 3d62be62e79f0f079eaa73e4072836da163da020 and also found one more and moved with 0eab7afe6ce0e2e14674b4a364faa52fdd75c37c ^^

src-tauri/src/constants.rs
```diff
@@ -7,3 +7,21 @@ pub const MASTER_SERVER_URL: &str = "https://northstar.tf";
 
 // server list endpoint
 pub const SERVER_BROWSER_ENDPOINT: &str = "/client/servers";
+
+// List of core Northstar mods
+pub const CORE_MODS: [&str; 3] = [
```
Alystrasz, line 12: I don't remember the context for #120, but this PR is the occasion to solve it :)

Alystrasz (changes requested):

src-tauri/src/constants.rs
GeckoEidechse: The context was that we declare core mods twice, once in frontend, once in backend which is just unnecessary duplication. As one cannot just do like 

```rust
#[ts(export)]
pub const BLACKLISTED_MODS: [&str; 3] = [
    "northstar-Northstar",
    "northstar-NorthstarReleaseCandidate",
    "ebkr-r2modman",
];
```

I'll have to add some function that frontend could call to get the list of core mods or other other constants. Will look into it but that shouldn't block this PR here, as this PR is basically a prerequisite for the other future one ^^

GeckoEidechse: Sorta addressed in #168

Alystrasz (approved):
LGTM.
Confirmed working on Windows.

PR #163, opened 2023-02-05
chore: Bump libthermite version to 0.5.2

GeckoEidechse:
And fix corresponding code that broke in the process

Basically just removing `await`s as the networking functions are no longer async.


Alystrasz (approved):
Confirmed working on Windows.

PR #165, opened 2023-02-05
feat: Change default resolution to SteamDeck size

GeckoEidechse:
This should give us fullscreen FlightCore out of the box when opening it in game UI. To my knowledge gamescope has no easy way of resizing window after opening.

This sidesteps it by just launching window in screen size.

As a nice side effect, this means by default more columns of mods are shown in the Thunderstore mod browser, previously it was just too small to fit 4 instead of 3 columns, making it look a bit awkward.

Before:
image

After:
image

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -64,9 +64,9 @@
         "resizable": true,
         "decorations": false,
         "title": "FlightCore",
-        "height": 600,
+        "height": 800,
         "minHeight": 300,
-        "width": 1000,
+        "width": 1280,
         "minWidth": 600
       }
     ]


Alystrasz:
A majority of FlightCore users don't use SteamDeck, so I think this is unneeded :)

Two suggestions:
* applying these dimensions only if current platform is SteamDeck;
* even better, saving dimensions in memory so they are restored on each launch.

GeckoEidechse:
> A majority of FlightCore users don't use SteamDeck, so I think this is unneeded :)
> 
> Two suggestions:
> 
> * applying these dimensions only if current platform is SteamDeck;
> 
> * even better, saving dimensions in memory so they are restored on each launch.

I can look into it. On [northstar_dev_testing_helper_tool](https://github.com/GeckoEidechse/northstar_dev_testing_helper_tool) I "solved" the SteamDeck detection by checking for Linux + `deck` username IIRC. Which works to some degree but doesn't take into consideration whether there might just be a desktop setup with username `deck`.

Also I still need to figure out setting window size in Tauri dynamically, cause right now I only know of the static method.

That being said, the previous windows size was chosen pretty arbitrarily and should at least increase the horizontal space by a bit in order to fit 4 columns of mods in the Thunderstore mod browser.

Cause the current look rn with just columns and lots of empty space just doesn't look good ^^

image

Alystrasz:
> Cause the current look rn with just columns and lots of empty space just doesn't look good ^^

I agree we should augment a bit the width to fit 4 columns; we shouldn't increase it too much, as the main view will feel even more empty.

> Also I still need to figure out setting window size in Tauri dynamically, cause right now I only know of the static method.

This should be straightforward to implement, using something like https://github.com/tauri-apps/tauri/discussions/3225#discussioncomment-2050864

GeckoEidechse:
> > Cause the current look rn with just columns and lots of empty space just doesn't look good ^^
> 
> I agree we should augment a bit the width to fit 4 columns; we shouldn't increase it too much, as the main view will feel even more empty.

Addressed in #169

> > Also I still need to figure out setting window size in Tauri dynamically, cause right now I only know of the static method.
> 
> This should be straightforward to implement, using something like [tauri-apps/tauri#3225 (reply in thread)](https://github.com/tauri-apps/tauri/discussions/3225#discussioncomment-2050864)

Investigating later when I get to it. Closing this PR for now ^^

PR #168, opened 2023-02-06
refactor: Get TS package API response from backend

GeckoEidechse:
Previously Thunderstore package index was done in frontend. Should be moved to backend instead as backend is reponsible for such tasks while frontend should just be used to store and display information.


GeckoEidechse:
@Alystrasz any idea how to get rid of the undefined warning?

image

Trying to use an if-guard just gives another error :(

image

GeckoEidechse:
CI failing cause Ubuntu Azure repos are down (still) and cause of TypeScript warning mentioned in https://github.com/R2NorthstarTools/FlightCore/pull/168#issuecomment-1419160472 that I will need assistance with to solve.

Alystrasz:
You need to initialize your `mods` variable, and type it as an array of Thunderstore mods.
Then, instead of checking if `mods` is undefined (which isn't a good practice btw), you can check if the list is empty:

```typescript
let mods: ThunderstoreMod[] = [];
await invoke<ThunderstoreMod[]>("query_thunderstore_packages_api")
    .then((message) => {
       mods = message;
    })
   .catch((error) => {
       console.error(error);
       return;
    });

if (mods.length === 0) {
    return;
}
```

GeckoEidechse:
> You need to initialize your `mods` variable, and type it as an array of Thunderstore mods.

Thanks for the clarification, fixed in 394f561114e74ce1ad693bf4702abe80008db494. Empty check isn't needed then anyway cause it's implicitly handled due it being an (empty) array. ^^

GeckoEidechse:
Assuming CI passes now, this PR should be done and ready for review ^^

GeckoEidechse:
Ok, now it should be fixed ^^"

Alystrasz (approved):
LGTM.
Confirmed working on Windows.

PR #173, opened 2023-02-08
feat: Add version number to user-agent

### LINKED ISSUE #170: Add version number to User-Agent
From pg:

> follow the same format as northstar, but use FlightCore instead of R2Northstar

GeckoEidechse:
Uses a crate that allows for formatting at compile time.

User agent is `FlightCore/MAJOR.MINOR.PATCH`, e.g. currently `FlightCore/1.7.0`.

Among other things, this allows for seeing percentage of different versions deployed on masterserver when FlightCore performs query to `/client/servers`.

Closes #170

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -92,6 +92,7 @@ dependencies = [
  "anyhow",
  "async-recursion",
  "chrono",
+ "const_format",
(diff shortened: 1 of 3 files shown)


Alystrasz (approved):
Confirmed working on Linux.

Issue #166, opened 2023-02-05
Figure out how to get proper stacktrace in release builds for Sentry


GeckoEidechse:
Currently stack traces in release builds on Sentry look like this

![image](https://user-images.githubusercontent.com/40122905/216847432-faefe454-dedf-4c7e-bab1-8ff48219c3ee.png)

Naturally this information is not very useful. At the very least the corresponding line in the source code would be nice to know to debug the crash.

GeckoEidechse:
Uploading corresponding pdb as suggested in Sentry docs has no effect as such I made an upstream issue now at https://github.com/getsentry/sentry/issues/45116

GeckoEidechse:
Debugged in:
- https://github.com/getsentry/sentry/issues/45116

Closed by:
- #192
- #195 
- #191

PR #175, opened 2023-02-08
feat: Better stacktrace in Sentry dashboard

GeckoEidechse:
Enable attaching stacktrace to see the function in which the crash happened.

https://docs.sentry.io/platforms/rust/configuration/options/#attach-stacktrace


GeckoEidechse:
Doesn't close #166 but at least gives more helpful info.

GeckoEidechse:
Not much to review here as you don't have access to the Sentry dashboard and it's a 1-line change anyway. Just put you up for review so that you are aware of it ig, @Alystrasz.

Otherwise could also just merge myself if need be ^^

PR #176, opened 2023-02-08
fix: Specify commit for tauri-plugin-store

GeckoEidechse:
Instead of a branch specify specific commit.
This is done to later run `cargo update`. Currently running said command updates the dependency to a broken build.

Fixing `tauri-plugin-store` to a specific working commit allows for running `cargo update` again.

Note that this is the same commit as the one specified in the `Cargo.lock` and incidentally also the last working commit of that crate ^^"


Alystrasz (approved):
Confirmed working on Windows.

PR #178, opened 2023-02-10
feat: Add ability to launch via Steam to DevView

Jan200101:
Some things will probably need to be overhauled


src-tauri/Cargo.toml
```diff
@@ -48,6 +48,8 @@ chrono = "0.4.23"
 ts-rs = "6.1"
 # const formatting
 const_format = "0.2.30"
+# open urls
+open = "3.2.0"
```
Jan200101, line 52: already a dependency pulled in by tauri

src-tauri/src/lib.rs
```diff
@@ -253,7 +253,7 @@ pub fn launch_northstar(
     let host_os = get_host_os();
 
     // Explicitly fail early certain (currently) unsupported install setups
-    if host_os != "windows"
+    if (host_os != "windows" && !matches!(game_install.install_type, InstallType::STEAM))
         || !(matches!(game_install.install_type, InstallType::STEAM)
             || matches!(game_install.install_type, InstallType::ORIGIN)
             || matches!(game_install.install_type, InstallType::UNKNOWN))
```
Jan200101, line 259: This is a mess but I don't know how this could be improved.

src-tauri/src/lib.rs
```diff
@@ -294,9 +297,28 @@ pub fn launch_northstar(
...
+        if host_os != "windows"
+        {
+            let mut file = std::fs::OpenOptions::new()
+                .create(true)
+                .write(true)
+                .truncate(true)
+                .open("run_northstar.txt").unwrap();
+
+            let mut content: &[u8] = b"1";
+            std::io::copy(&mut content, &mut file).unwrap();
```
Jan200101, line 312: This is copied from `extract`, there are many ways to do file operations in Rust and none of them appear to be that great.

Error handling could be improved to explicitly state why we bailed.

src-tauri/src/lib.rs
```diff
@@ -294,9 +297,28 @@ pub fn launch_northstar(
...
+        }
+
+        return match open::that("steam://run/1237970/") {
+            Ok(()) => Ok("Launched game".to_string()),
+            Err(_err) => Err(anyhow!("Failed to launch Titanfall 2 via Steam").to_string()),
+        }
```
Jan200101, line 318: Should work on any System where Steam handles its own scheme (Tested on Windows and Fedora 37).

Another implementation (that would not add another explicit dependency) would be finding Steam (via steamlocate) and executing executing `steam.exe`/`steam.sh` with that.

src-tauri/src/lib.rs
GeckoEidechse: Eh if it works it's alright. Wouldn't be the most jank code in the FlightCore code base :P

What I'm wondering more is if we should maybe have a post run call somewhere in FlightCore that cleans up `run_northstar.txt` again if the file wasn't there in the first place?

Otherwise we would have the side-effect of (unintentionally) causing the player to always launch Northstar when they launch Titanfall2 ^^

src-tauri/src/lib.rs
GeckoEidechse: Might need someone to confirm whether Steam Flatpak will also work properly with URI. I'll ask around in the Linux channel on the Northstar Discord.

Personally I was planning to only use URIs too once I'd get around to implementing Linux launch support. So unless it turns out that it decent work for a decent chunk of Linux users, I'm fine with doing it just via URIs ^^

GeckoEidechse:
To fix the failing format check, either use autoformatting functiona

GeckoEidechse:
To fix failing format check, either use auto-format functionality of your IDE on `lib.rs` or run `cargo fmt --manifest-path src-tauri/Cargo.toml --all` in repo root and commit the changes in `lib.rs` ^^

src-tauri/src/lib.rs
GeckoEidechse: So far seems like it works for people in the Linux channel ^^

src-tauri/src/lib.rs
```diff
@@ -294,9 +297,28 @@ pub fn launch_northstar(
     }
 
     // Only Windows with Steam or Origin are supported at the moment
-    if host_os == "windows"
-        && (matches!(game_install.install_type, InstallType::STEAM)
-            || matches!(game_install.install_type, InstallType::ORIGIN)
+    if matches!(game_install.install_type, InstallType::STEAM)
+    {
+        // Create run_northstar.txt for NorthstarProton
+        if host_os != "windows"
```
Jan200101, line 303: Should `run_northstar.txt` also be used on Windows?
The codepath exists is only required on Linux to work around some issues (afaik)-

src-tauri/src/lib.rs
Jan200101: This should 100% be done.

src-tauri/src/lib.rs
```diff
@@ -294,9 +297,28 @@ pub fn launch_northstar(
...
+                .create(true)
+                .write(true)
+                .truncate(true)
+                .open("run_northstar.txt").unwrap();
+
+            let mut content: &[u8] = b"1";
+            std::io::copy(&mut content, &mut file).unwrap();
+        }
+
+        return match open::that("steam://run/1237970/") {
```
Jan200101, line 315: right now no arguments are passed, meaning it uses whatever arguments were set up on Steam.

Should this be kept, so that the launch arguments the user decided on are used, or should `-northstar` be enforced?

src-tauri/src/lib.rs
```diff
@@ -277,12 +277,15 @@ pub fn launch_northstar(
...
+        // Require Origin to be running to launch Northstar except for Steam
+        if !matches!(game_install.install_type, InstallType::STEAM)
+        {
+            let origin_is_running = check_origin_running();
+            if !origin_is_running {
+                return Err(
+                    anyhow!("Origin not running, start Origin before launching Northstar").to_string(),
+                );
+            }
         }
```
Jan200101, line 289: Should this be moved to the code below that runs `NorthstarLauncher.exe`?

src-tauri/src/lib.rs
Jan200101, line 315: Another thing that needs to be considered is NorthstarProton on Linux.

Steam only parses available Proton versions on startup, and will ignore modification to `config.vdf`.

This means we need Steam closed and then restarted (or at least instruct the user on how to set it up)

GeckoEidechse: If you do not use `run_northstar.txt` on Windows, you'd need to pass `-northstar`, otherwise it'll launch vanilla Titanfall2. Not sure if there was a Steam URI that allowed launching with params. If not we might have to use `run_northstar.txt` on Windows as well.

GeckoEidechse: Hmm, I really wish there was a way to tell Steam to use a specific Proton version for a single run of a game. Killing and restarting Steam on every launch really doesn't sound like a good idea ^^

src-tauri/src/lib.rs
GeckoEidechse: Given that `-northstar` doesn't work a for a reason that is still unknown, it might be simpler to just use `run_northstar.txt` on both Windows and Linux. Otherwise you have to use `-northstar` on one and `run_northstar.txt` on the other ^^

src-tauri/src/lib.rs
GeckoEidechse: Imma make a new issue to maybe figure out control flow for launching Northstar ^^

GeckoEidechse: Got a potential flow chart on how to structure the control flow in https://github.com/R2NorthstarTools/FlightCore/issues/179#issuecomment-1426895213

Feel free to leave any comments/concerns/ideas/etc there.

Question is, do we wanna change control flow as part of this PR or figure it out in a separate one? ^^

src-tauri/src/lib.rs
Jan200101: I don't think we can do that either way, we cannot control how the user set up Steam.

Best we can do is setup NorthstarProton (libthermite provides the capbilities, we just need to do it) and then check if NorthstarProton is the compatibility tool setup for Titanfall.

The user would only need to be instructed to restart Steam and change Proton and be informed when Titanfall isn't setup to use the right Proton.

Jan200101: You can launch a game with arguments via the browser protocol: `steam://run/<id>//<args>/`

but it gives the user a prompt showing them what arguments were passed
image

This ruins the UX so I think relying on `run_northstar.txt` is the way to go here.

GeckoEidechse: Agreed :D

(Also kinda glad Steam has that check ^^)

src-tauri/src/lib.rs
Jan200101: Okay, I've been implementing this and...

when do we clean up?
We don't control the launching and just tell Steam to do it.
Do we wait for Titanfall2.exe to launch and then move the files back?
Or maybe for Titanfall2.exe to exit? but then we need to persist after the games launch.

GeckoEidechse: On a sidenote, given that control is starting to get messy (and needs restructuring anyway, looking at #179), it might make sense to just move all the code for Steam launching into a dedicated function and have it hooked up to a button in DeveloperView.

Then we can just ship that in an update and have people test it that way.

Then we refactor the launching process and from there incorporate the function for launching via Steam. ^^

Jan200101: > I'd change it back after `NorthstarLauncher.exe` was detected as running.

We don't run `NorthstarLauncher.exe` on Linux (or Windows when launching over Steam).
We cannot replace `Titanfall2.exe` because that would make the install unclean possible causing future updates to reset it.
The launch parameters cannot be hijacked to launch `NorthstarLauncher.exe` either because its setup to interact with EA's App directly (see [SteamDB](https://steamdb.info/app/1237970/config/))

So there only really is `Titanfall2.exe`
There might be some way of checking if the running Titanfall process is Northstar by seeing what is loaded, but I think that is a bit overkill.
It is fair to assume any Titanfall2.exe or NorthstarLauncher.exe might be Northstar, in which case launching should be prevented.

> Cause at that point the wsock32 proxy did its job and launched Northstar.

unsure.
wsock32 is a runtime dependency and there is no reliable duration it will take for it to load the configuration as it can depend on multiple factors (e.g. computer hardware)

Inotify can detect when a file was opened, that could be used to check when Northstar read the file, its possibly something similar exists for Windows

> Only issue is when user closes FlightCore before Northstar launched. Don't really have a solution for that one tbh...

While I agree that it would be best to clean up after ourselves and bring the environment back into the state we found it in, maybe there is no good way of doing this.

GeckoEidechse: > We don't run `NorthstarLauncher.exe` on Linux (or Windows when launching over Steam).

Oh I thought it starts it as a new process nevermind then my mistake ^^

Will require some testing but it might be enough to check for a running `Titanfall2.exe` then and after X seconds start the clean-up process.

Jan200101:
Separated the Steam Launching code into its own function that can be called via a button in the Dev menu.

On Systems that are not Windows it checks for NorthstarProton by parsing `config.vdf` using keyvalues-parser, steamy-vdf is unmaintained and cannot parse that file because it parses VDFs incorrectly.

The cleanup is done in the background with a 3 minute wait for Titanfall to launch.
3 Minutes is the longest my laptop took to Launch the EA App and for that to launch Titanfall, though it might still go wrong.

Some things are still done in rather dirty way and would break if the user creates a very specific directory structure, but otherwise it works.

src-tauri/src/lib.rs
```diff
@@ -314,6 +317,99 @@ pub fn launch_northstar(
...
+
+    let vdf_text = fs::read_to_string(configpath).unwrap();
+    let config = Vdf::parse(&vdf_text).unwrap().value;
+    let obj = config.unwrap_obj();
+
+    let software = obj.get("Software")?.get(0)?.get_obj()?;
+    let valve = software.get("Valve")?.get(0)?.get_obj()?;
+    let steam = valve.get("Steam")?.get(0)?.get_obj()?;
+    let compat_tool_mapping = steam.get("CompatToolMapping")?.get(0)?.get_obj()?;
+    let tfapp = compat_tool_mapping.get("1237970")?.get(0)?.get_obj()?;
```
GeckoEidechse, line 332: Not sure about other values but I'd move the Steam app id for Titanfall2 to `constants.rs` at least, that way you can pull it in other locations (e.g. https://github.com/R2NorthstarTools/FlightCore/pull/178/files#diff-eabcebd0ae5c9b77a5247e1bdbdb88b1869fc0c932d0ef0cf2ecf9ee5221be7cR382) as well ^^

GeckoEidechse (changes requested):
Aight, first look at it again. Will do some testing later when I get to it ^^

src-tauri/src/lib.rs
GeckoEidechse: Marking as _resolved_ as no longer relevant ^^

src-tauri/src/lib.rs
GeckoEidechse: Marking as _resolved_ as no longer relevant for this PR ^^

GeckoEidechse:
> Will do some testing later when I get to it ^^

Ugh, EA App currently fails when I try to launch Titanfall2 via Steam (vanilla or not, unrelated to the PR) T_T

GeckoEidechse:
Confirmed working on Windows :D

src-tauri/src/lib.rs
```diff
@@ -314,6 +319,97 @@ pub fn launch_northstar(
     ))
 }
 
+pub fn get_titanfall_proton() -> Option<String> {
```
GeckoEidechse, line 322: Maybe add a [doc comment](https://doc.rust-lang.org/reference/comments.html#doc-comments) (`///`) that shortly explains what the function does ^^

src-tauri/src/lib.rs
```diff
@@ -314,6 +319,97 @@ pub fn launch_northstar(
...
+
+    let software = obj.get("Software")?.get(0)?.get_obj()?;
+    let valve = software.get("Valve")?.get(0)?.get_obj()?;
+    let steam = valve.get("Steam")?.get(0)?.get_obj()?;
+    let compat_tool_mapping = steam.get("CompatToolMapping")?.get(0)?.get_obj()?;
+    let tfapp = compat_tool_mapping
+        .get(TITANFALL2_STEAM_ID)?
+        .get(0)?
+        .get_obj()?;
+    let name = tfapp.get("name")?.get(0)?.get_str()?.to_string();
```
GeckoEidechse, line 338: This whole parsing mess is cause the steam vdf parsing crate we use doesn't support parsing for proton version, right?

src-tauri/src/lib.rs
```diff
@@ -314,6 +319,97 @@ pub fn launch_northstar(
...
+        // We failed to get to Titanfall2 directory
+        return Err("Couldn't access Titanfall2 directory".to_string());
+    }
+
+    let run_northstar = "run_northstar.txt";
+    let run_northstar_bak = "run_northstar.txt.bak";
+
+    if Path::new(run_northstar).exists() {
+        // rename should ovewrite existing files
+        // TODO: deal with folders
```
GeckoEidechse, line 375: How exactly do you need to deal with folders? ^^

src-tauri/src/lib.rs
```diff
@@ -314,6 +319,97 @@ pub fn launch_northstar(
...
+        }
+
+        // Northstar may be running, but it may not have loaded the file yet
+        sleep(Duration::from_millis(2000)).await;
+
+        // intentionally ignore Result
+        let _ = fs::remove_file(run_northstar);
+
+        if Path::new(run_northstar_bak).exists() {
+            // TODO: deal with folders
```
GeckoEidechse, line 405: How exactly do you need to deal with folders? ^^

src-tauri/src/lib.rs
```diff
@@ -314,6 +319,97 @@ pub fn launch_northstar(
...
+    let tfapp = compat_tool_mapping
+        .get(TITANFALL2_STEAM_ID)?
+        .get(0)?
+        .get_obj()?;
+    let name = tfapp.get("name")?.get(0)?.get_str()?.to_string();
+
+    return Some(name);
+}
+
+pub fn launch_northstar_steam(
```
GeckoEidechse, line 343: Maybe add a [doc comment](https://doc.rust-lang.org/reference/comments.html#doc-comments) (`///`) that shortly explains what the function does ^^

GeckoEidechse (commented):
Working on Windows on my end. Code also looks good so far. A few smaller corrections but then we should be able to merge and roll out for wider testing IMO ^^

src-tauri/src/lib.rs
Jan200101: Yes.

I can go and submit a PR to `steamlocate-rs` to add a helper function to fetch the compatibility tool for a given appid if that desired.

src-tauri/src/lib.rs
Jan200101: This is in regards to the implementation details of `fs::rename`.

> On Unix, if from is a directory, to must also be an (empty) directory. If from is not a directory, to must also be not a directory.

so, to be sane, it would be best to check what `run_northstar.txt` and its backup actually are.

src-tauri/src/lib.rs
GeckoEidechse: Think this is resolved by now ^^

src-tauri/src/lib.rs
GeckoEidechse: I mean, I'm not gonna stop you if you want to ;)

src-tauri/src/lib.rs
GeckoEidechse: Ah so one could have a folder called `run_northstar.txt/` and then the function would actually fail.

Tbh, that's rather unlikely scenario and I'm more than happy to ignore it. If the function just panics in that case I could use Sentry dashboard to see if  / how often we hit it and apply a workaround if need be ^^

Jan200101: Alright, I'll remove the comments when I find the time.

src-tauri/src/lib.rs
Jan200101: https://github.com/WilliamVenner/steamlocate-rs/pull/18

Jan200101: I don't think the steamlocate-rs PR will be merged any time soon.

GeckoEidechse: Yeah, I guess we'll just have to keep the messy part in FlightCore then ^^"

GeckoEidechse: Should the PR get merged we can always strip it out again.

src-tauri/src/lib.rs
```diff
@@ -365,7 +467,7 @@ pub fn get_enabled_mods(game_install: GameInstall) -> Result<serde_json::value::
     let enabledmods_json_path = format!("{}/R2Northstar/enabledmods.json", game_install.game_path);
 
     // Check for JSON file
-    if !std::path::Path::new(&enabledmods_json_path).exists() {
+    if !Path::new(&enabledmods_json_path).exists() {
```
GeckoEidechse, line 470: Fixups should be separate PRs just to keep the commit diff contained ^^

src-tauri/src/lib.rs
GeckoEidechse: Still valid btw :P

src-tauri/src/lib.rs
GeckoEidechse: Reverted in 9218b254dffbf58dd78f96666eb7d457519f1c76

GeckoEidechse:
Ugh, looks like another PR that just got merged introduced merge conflicts again. Sorry about that ^^"

Jan200101:
merged master into this to fix any conflicts
migrated to a newer steamlocate version which has the compat_tool code.

src-tauri/Cargo.toml
```diff
@@ -29,7 +29,7 @@ tokio = { version = "1", features = ["full"] }
 sentry = "0.30"
 sentry-log = "0.30.0"
 # Find steam games
-steamlocate = "1.0.2"
+steamlocate = "1.2.*"
```
GeckoEidechse, line 32: Bit nit-picky but the recommended syntax to just pick newest patch level is just to omit it, i.e.

```suggestion
steamlocate = "1.2"
```

src-tauri/src/lib.rs
```diff
@@ -328,6 +331,93 @@ pub fn launch_northstar(
...
+        return Err("Couldn't access Titanfall2 directory".to_string());
+    }
+
+    let run_northstar = "run_northstar.txt";
+    let run_northstar_bak = "run_northstar.txt.bak";
+
+    if Path::new(run_northstar).exists() {
+        // rename should ovewrite existing files
+        fs::rename(run_northstar, run_northstar_bak).unwrap();
+    }
```
GeckoEidechse, line 386: I just realised, spamming _launch via Steam_ will basically overwrite backup with the `1` version. Tbh, not a really a big issue but just something to keep in mind.

Might gonna later add a function to repair window that allows easily removing the file again ^^

src-tauri/src/lib.rs
```diff
@@ -328,6 +331,93 @@ pub fn launch_northstar(
...
+
+    // Switch to Titanfall2 directory to set everything up
+    if std::env::set_current_dir(game_install.game_path.clone()).is_err() {
+        // We failed to get to Titanfall2 directory
+        return Err("Couldn't access Titanfall2 directory".to_string());
+    }
+
+    let run_northstar = "run_northstar.txt";
+    let run_northstar_bak = "run_northstar.txt.bak";
```
GeckoEidechse, line 381: Out of curiosity, is `.bak` an actual convention file ending for backup files? ^^

GeckoEidechse (commented):
Few minor things, code looks good otherwise. Will test tonight hopefully and merge if all good :D

src-tauri/src/lib.rs
Jan200101: https://en.wikipedia.org/wiki/Bak_file
alternatives include but are not limited to
`file~`
`file.orig`
`file.old`

GeckoEidechse: Yeah, that's fine then. Was just curious cause I wasn't aware of any convention ^^

GeckoEidechse (approved):
(Re-)Confirmed working on Windows :D

GeckoEidechse:
Testing from complete cold start (both Steam and EA App closed) on Windows and it just launched vanilla instead now? On warm rerun it launched Northstar again. Maybe the `run_northstar.txt` is restored back too soon?

Jan200101:
> Testing from complete cold start (both Steam and EA App closed) on Windows and it just launched vanilla instead now? On warm rerun it launched Northstar again. Maybe the `run_northstar.txt` is restored back too soon?

question is, how long should we wait, or should we wait at all?

Titanfall may not launch because of a crash, so we can't wait for it until the heat death of the universe.
Longer delays make it more likely that a user will prevent the cleanup by closing FlightCore.
Shorter delays make it more likely that Titanfall hasn't even started on weaker hardware and cold starts when we run the cleanup.

There is also the thing that Steam could not be running and will take a bit to start up.
Or that the user isn't logged in.
Or that the user doesn't have Origin installed
Or that the user isn't logged into Origin
Or that Origin is taking a while to load

It is truly a cat and mouse game.

GeckoEidechse:
> > Testing from complete cold start (both Steam and EA App closed) on Windows and it just launched vanilla instead now? On warm rerun it launched Northstar again. Maybe the `run_northstar.txt` is restored back too soon?
> 
> question is, how long should we wait, or should we wait at all?
> 

I think best we can do is heuristic based. So like spawn a process on the side that checks when/if Steam/EA-App launched and then based on that reduces/changes the countdown timer so that when for someone Steam launches within a few seconds the cooldown timer is lower than for someone where it might take a minute for Steam to launch ^^

That being said, this is kinda outside the scope of this PR and Steam launch is still gated behind DevView rn anyway so it doesn't have to be perfect ^^

src-tauri/Cargo.toml
GeckoEidechse: Adding fix myself

PR #180, opened 2023-02-12
feat: Install NorthstarProton

GeckoEidechse:
This branch is still very much WIP. Still need to figure a few things out. Expect force pushes to clean up commits!


GeckoEidechse:
I'll fix the build issues on Windows later today but after that I think I won't be able to work on this PR for a bit. So if anyone else was considering working on it, please do <3

GeckoEidechse:
Replaced by #383

PR #186, opened 2023-02-22
refactor: Properly set cfg macro for OS conditional compilation

GeckoEidechse:
Makes sure that Windows/Linux specific functions are only compiled on their respective OS.

Will also help with reducing diff of #180

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -9,6 +9,7 @@ mod platform_specific;
 #[cfg(target_os = "windows")]
 use platform_specific::windows;
 
+#[cfg(target_os = "linux")]
 use platform_specific::linux;
 
 use serde::{Deserialize, Serialize};
@@ -67,7 +68,7 @@ pub fn check_mod_version_number(path_to_mod_folder: String) -> Result<String, an
 // I intend to add more linux related stuff to check here, so making a func
 // for now tho it only checks `ldd --version`
 // - salmon
-
+#[cfg(target_os = "linux")]
 pub fn linux_checks_librs() -> Result<(), String> {
     // Perform various checks in terms of Linux compatibility
     // Return early with error message if a check fails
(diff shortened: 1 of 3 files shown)


GeckoEidechse:
Note that if MacOS compilation wasn't broken by now this will definitely break it due to

https://github.com/R2NorthstarTools/FlightCore/pull/186/files#diff-2f5e0a90d4195e9986f5e24928dce16b59a80a2cf30f7059b38d55bd7d1eff69L145-R154

having explicit cfg macros for Windows/Linux but not for "other"/MacOS. Not that it matters cause we never targetted and not hard to fix either but just something to be aware of ig. ^^

GeckoEidechse:
(just noticed I assigned instead of requesting review again, whoops. Sorry @Alystrasz ^^")

Alystrasz:
> Note that if MacOS compilation wasn't broken by now this will definitely break it due to
> 
> https://github.com/R2NorthstarTools/FlightCore/pull/186/files#diff-2f5e0a90d4195e9986f5e24928dce16b59a80a2cf30f7059b38d55bd7d1eff69L145-R154
> 
> having explicit cfg macros for Windows/Linux but not for "other"/MacOS. Not that it matters cause we never targetted and not hard to fix either but just something to be aware of ig. ^^

Can you write a comment somewhere so we don't forget this?

GeckoEidechse:
> Can you write a comment somewhere so we don't forget this?

Done in 14945ee65150f3e6c2f6bb659551b9dd848c60cc

Alystrasz (commented):
I do have the error message on Windows, gotta try with Ubuntu.

GeckoEidechse:
> I do have the error message on Windows, gotta try with Ubuntu.

What error message? ^^"

Alystrasz:
> > I do have the error message on Windows, gotta try with Ubuntu.
> 
> What error message? ^^"

"Not available on Windows"

Alystrasz (approved):
LGTM.
Works on both Windows and Linux.

PR #187, opened 2023-02-23
build: Separate zips for artifacts for Windows and Linux

GeckoEidechse:
Splits up the Windows and Linux artifacts into separate zip folders.

This is especially useful for testing Windows builds as the Linux ones are huge due to being an AppImage. Windows is tiny (7MB) in comparison and as such this means when testing CI build on Windows with this PR one goes from downloading 300MB (Windows + Linux files) to just 7MB (only Windows files).

image


Alystrasz (approved):
Confirmed working through CI build.

PR #184, opened 2023-02-18
feat: Generate Northstar release notes

GeckoEidechse:
Allows generating release notes for Northstar within FlightCore

This is basically a developer tool to make my life easier but should also help any other potential future Northstar release manager.

TODO:
- [ ] code cleanup

PR #188, opened 2023-02-23
feat: Generate FlightCore release notes

GeckoEidechse:
Allows generating release notes for FlightCore within FlightCore 🤯 

This is basically a developer tool to make my life easier but in turn that means I'm able to push out releases for FlightCore faster as another piece of friction is removed ^^

It's similar to #184 but for FlightCore (combining the two PRs will be fun :>)

TODOs
- [x] Handle commit types not in sorted type list
- [x] Code cleanup
- [x] https://github.com/R2NorthstarTools/FlightCore/pull/188#discussion_r1121887167


src-vue/src/views/DeveloperView.vue
```diff
@@ -4,6 +4,39 @@
...
+            <el-button type="primary" @click="getTags">
+                Get tags
+            </el-button>
+
+            <el-select v-model="firstTag" class="m-2" placeholder="First tag">
+                <el-option
+                    v-for="item in ns_release_tags"
+                    :key="item.name"
+                    :label="item.name"
+                    :value="item.name"
```
GeckoEidechse, line 17: I cannot get it to sort list by `item.name` but on click have `item` be the selected object (as opposed to `item.name`).

Any help on that would be appreciated <3

Alystrasz: Why do you want to do this? Do you want to put some stuff inside your `item` instances?

GeckoEidechse: The item is of type [tag](https://github.com/R2NorthstarTools/FlightCore/pull/188/files#diff-adfc12dc30d4e3e06e9a49d216f3db3ca6cf3b493dc344db7793fb814a48d8ebR3) which right now only has a single member called `name`. For the future, I might also consider adding information to the type concerning repo or commit hash.

Naturally this doesn't need to be shown in the dropdown menu but I still need to pass the full object to [`compareTags`](https://github.com/R2NorthstarTools/FlightCore/pull/188/files#diff-0944e810871d55931e96ce560916f7fe1048c4a18cfbd3ab18a8a83dd59c968aR300-R317) which in turn passes it to the backend function.

And I cannot figure out how to get the list sorted by `tag.name` but on selecting an element of the list set the value of [`firstTag`](https://github.com/R2NorthstarTools/FlightCore/pull/188/files#diff-0944e810871d55931e96ce560916f7fe1048c4a18cfbd3ab18a8a83dd59c968aR127-R134) and [`secondTag`](https://github.com/R2NorthstarTools/FlightCore/pull/188/files#diff-0944e810871d55931e96ce560916f7fe1048c4a18cfbd3ab18a8a83dd59c968aR135-R141) to the full `tag` object, not just the content of the `name` field... ^^"

Alystrasz: [According to the documentation](https://element-plus.org/en-US/component/select.html#basic-usage), selector values must be objects having two keys only, `label` and `value`.
Maybe you can put `ns_release_tags` objects as `value`s?

Alystrasz: @GeckoEidechse did you manage to solve this issue?

GeckoEidechse: I attempted, saw that it can work but needs some refactoring, and then got busy with something else ^^"

GeckoEidechse: Should now be fixed with 5f636608ee44bce09e7778115ac07877fb5d978a and 313585e952f3245cda7a2a8c97bf7dcc29a7cf2a ^^

src-vue/src/views/DeveloperView.vue
```diff
@@ -152,6 +210,42 @@ export default defineComponent({
                     });
                 });
         },
+        async getTags() {
+            await invoke<TagWrapper[]>("get_list_of_tags")
+                .then((message) => {
+                    this.ns_release_tags = message;
+                    ElNotification({
```
Alystrasz, line 217: These empty notifications feel weird.
Maybe you can put a success message there?

src-tauri/src/github/mod.rs
```diff
@@ -1,2 +1,179 @@
...
+#[derive(Debug, Deserialize)]
+struct Comparison {
+    commits: Vec<CommitInfo>,
+}
+
+/// Get a list of tags on the FlightCore repo
+#[tauri::command]
+pub fn get_list_of_tags() -> Result<Vec<TagWrapper>, String> {
+    // Set the repository name.
+    let repo = "R2NorthstarTools/FlightCore";
```
Alystrasz, line 43: Shouldn't this be in constants package?

GeckoEidechse: Good catch, done in ee016b85d2d4b53c0ef1f0822bbd6b0e47f30ac7

src-vue/src/views/DeveloperView.vue
```diff
@@ -63,8 +96,33 @@ export default defineComponent({
...
         return {
             mod_to_install_field_string : "",
+            release_notes_text : "",
+            first_tag: { label: '', value: {name: ''} },
+            second_tag: { label: '', value: {name: ''} },
+            ns_release_tags: [] as TagWrapper[],
         }
     },
+    computed: {
+        ns_release_tags(): TagWrapper[] {
```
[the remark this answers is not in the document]
GeckoEidechse: Cause I still kinda don't quite understand how Vue.js works and just copy/paste stuff together ^^"

Anyway, removed it in 4e77370022a1c921fddf8302795d2e1d575da32e and seems to still work

Alystrasz (approved):
Confirmed working on Windows.

PR #190, opened 2023-02-25
build: Add Volar to recommended vscode extensions

GeckoEidechse:
Adds `vue.Volar` to list of recommended vscode extensions. This should have been there already from the get-go tbh ^^


Alystrasz (approved):
I didn't know this extension since I don't usually use vscode.
Nice addition though :)

PR #191, opened 2023-02-25
build: Keep line tables in release builds

GeckoEidechse:
This allows for seeing line numbers on crash logs from release builds on sentry.io

Right now my stack traces on Sentry look like this which makes it incredibly difficult to figure out the root issue.

image

Seeing line numbers naturally simplifies the debugging process by a longshot. This comes at the cost of increased binary size on Linux (Windows stores debug symbols in separate file) but is worth the cost as it allows being able to percisely debug any crashes.

Requires:
- [x] #195


GeckoEidechse:
Line numbers show up properly when testing on Linux :D 

image

GeckoEidechse:
Not working on Windows, ugh

image

GeckoEidechse:
Confirmed working on two separate devices :D

image

PR #192, opened 2023-02-26
Upload Windows `.pdb`

GeckoEidechse:
Upload Windows debug symbols which are stored in `.pdb` file to artifacts.

This is done as part of an attempt to solve #166


GeckoEidechse:
Merging myself cause this is only a 2-line CI change.

PR #195, opened 2023-02-27
chore: Bump sentry to 0.30.0

GeckoEidechse:
Update `sentry` backend dependency to newest release. Solved #166 as on the newer release `debug_images` are uploaded by default.

Note that this raises the min Rust version required to `1.66`. If that's an issue we could also use `0.29.3` of `sentry` but I don't really see a point to purposefully stay on older release ^^

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -914,6 +914,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "findshlibs"
+version = "0.10.2"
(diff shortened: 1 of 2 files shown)


GeckoEidechse:
As discussed in https://github.com/getsentry/sentry/issues/45116 this solves #166 (I still need to manually upload `pdb`s but I can solve that in a separate issue :P)

GeckoEidechse:
Merging myself as small change (only single dependency update) yet high priority

Issue #189, opened 2023-02-24
Update used Actions in CI


GeckoEidechse:
We still use some older versions that are to be deprecated:

![image](https://user-images.githubusercontent.com/40122905/221203927-d42b7967-abc2-4463-b2dc-2ecc11cb3323.png)

~~Requires #187 to be merged first as otherwise I'm gonna have a merge conflict ^^"~~

**Progress:**
- [x] #196
- [ ] #210 
- [ ] Update any remaining actions

GeckoEidechse:
Seems like we might have an issue: https://github.com/actions-rs/toolchain/issues/221

Action we use seems to be unmaintained. Naturally we're not the only ones using it so I guess it's wait and see whether someone will step up in time or whether we'll just have to go back to manually installing rust + cargo first...

GeckoEidechse:
https://github.com/dtolnay/rust-toolchain seems to be the recommended replacement according to reddit posts on the subject.

- https://old.reddit.com/r/rust/comments/vyx4oj/actionsrs_organization_became_unmaintained/ig54zv7/
- https://old.reddit.com/r/rust/comments/z1mlls/actionsrs_github_actions_need_more_maintainers_or/ixbwr26/

PR #196, opened 2023-02-27
chore: Bump checkout action to v3

GeckoEidechse:
v2 used an outdated version of Node JS and GitHub warns about deprecation. As such we should bump to newer release.

Part of solving
- #189


Alystrasz (approved):
All CI checks are green, so I don't see any reasons not to merge this :)

PR #197, opened 2023-02-28
feat: "Install PR" UI

Alystrasz:
Little improvements over the PR installation feature:

* PRs are fetched on collapse item click;
* Loaders indicate PRs are being fetched.

I could improve things more, but since it's a developer feature, it doesn't need more polish imo :)

Capture d’écran de 2023-02-28 08-21-35

### DIFF
--- a/src-vue/src/components/PullRequestsSelector.vue
+++ b/src-vue/src/components/PullRequestsSelector.vue
@@ -1,6 +1,6 @@
 <template>
     <div>
-        <el-collapse>
+        <el-collapse accordion>
             <el-collapse-item title="Launcher PRs" name="1">
                 <el-button type="primary" @click="getPullRequests('LAUNCHER')">
                     Get launcher PRs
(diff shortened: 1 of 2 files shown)


src-vue/src/components/PullRequestsSelector.vue
```diff
@@ -0,0 +1,184 @@
+<template>
+    <div>
+        <el-collapse accordion @change="onChange">
```
GeckoEidechse, line 3:
```suggestion
        <el-collapse @change="onChange">
```

I'd remove the accordion attribute. When I need to look at pull request that needs both Mods and Launcher, it's a bit too much friction to constantly have to open the other collapsed list again ^^

I'm aware with both open the list gets a bit cluttered, hence we probably wanna add a search feature for each repo next that could be used to then filter to the desired PR but that's outside of the scope of this PR :P

GeckoEidechse (commented):
From quick test everything still works as expected. Love the load on collapse open btw <3

The only thing I'd change is removing the accordion attribute, then it should be good to merge ^^

src-vue/src/components/PullRequestsSelector.vue
```diff
@@ -0,0 +1,184 @@
+<template>
+    <div>
+        <el-collapse @change="onChange">
```
GeckoEidechse, line 3: Ok turns out removing the `accordion` attribute breaks calling `getPullRequests()`? Wtf?

Didn't happen when testing before previous review cause it had already loaded the list of PRs and the hot-reloaded when I removed the attribute.

Running 6ab48c0f3ec21fb8844a8a18b26146546ae1ea4a from cold start I cannot get the PR list to load for neither Mods nor Launcher. Adding a debug `console.log()` at the beginning of `getPullRequests()` I can tell that the function is never called x_x

GeckoEidechse: Confirmed fixed in 27a0e892dcaf6a88a93f8679993854c8a915a99e

GeckoEidechse (approved):
Confirmed working as intended now :D

PR #198, opened 2023-03-01
chore: Bump all dependencies

GeckoEidechse:
Bumps all dependencies. Most notably updates Tauri to `v1.2.3` from `v1.2.1`


Alystrasz (approved):
I confirm both development and release builds work with these modifications (using Windows at least).

PR #199, opened 2023-03-03
chore: Bump libthermite to 0.5.3

GeckoEidechse:
This adds a change needed for making progressbar work on Northstar/mod download.


Alystrasz (approved):
Confirmed running and building fine on Windows.

PR #200, opened 2023-03-03
feat: Show download progress for installing Northstar

GeckoEidechse:
Still WIP. The end goal would be to have a way to show download and install progress of Northstar.

Right now we do not show any progress which is frustrating to anyone where the install takes longer than a few seconds.

TODO:
- [x] i18n


GeckoEidechse:
Currently this is done using events which is fine but I need to pass the window handle all the way down to the install function which is annoying. If I could somehow instead grab the default window or something similar that doesn't require me to pass the window object all the way down, that would be great ^^

GeckoEidechse:
Okay, I think I got the basics figured out. From here I "only" need to refactor and settle on a format for transmitting data...

@Alystrasz I assume I just transfer the raw info, i.e. downloaded bytes and total size and then do the calculation on the frontend?

Alystrasz:
> @Alystrasz I assume I just transfer the raw info, i.e. downloaded bytes and total size and then do the calculation on the frontend?

Yep you only need those two values to display download progress in UI :)

Do you plan to distinguish between archive downloading and extraction?
(I had the issue with mods autodownload where extraction would take more time than downloading)

GeckoEidechse:
> Do you plan to distinguish between archive downloading and extraction? (I had the issue with mods autodownload where extraction would take more time than downloading)

Yes. I guess the way to go would be to also supply an enum in the message that indicates the state i.e. `DOWNLOADING` vs `EXTRACTING`? ^^

GeckoEidechse:
> > Yes. I guess the way to go would be to also supply an enum in the message that indicates the state i.e. `DOWNLOADING` vs `EXTRACTING`? ^^
> 
> That's what I ended up doing, yep

Alright, should be implemented now. For now it simply does a `console.log`. I suppose you'd be willing to take it from here and look at the front-end part of it? ^^"

image

Alystrasz:
Why do you only have 3 download updates?
No extraction updates at all? :(

GeckoEidechse:
> Why do you only have 3 download updates?

It sends an update every 250ms. However if you have a 1gbps downlink, that means that those 70MB of Northstar zip get downloaded in less than a second, i.e. only 3 updates sent out before download is done xD

> No extraction updates at all? :(

Don't know yet how to do progress tracking of the extraction unfortunately. Need to learn more Rust before I do but that's gonna take a few weeks based on how much time I can invest into this atm :/

GeckoEidechse:
Tried a bit with implementing a progress bar myself but I'm failing on actually listening to the event in the right Vue component... ^^"

Alystrasz:
Listeners added from your component do not receive pings from backend?

GeckoEidechse:
> Listeners added from your component do not receive pings from backend?

No, not knowing exactly what to do I thought I could just use ChatGPT but the answer it gave me was super convoluted. I didn't realise I could've just added the event listener in the front-end...

GeckoEidechse:
Okay, doesn't look super nice but I got a working progressbar now :D

GeckoEidechse:
Aight, I got some rudimentary progress bar done now. Only issue is that CSS seems to be squishing it to the point that it becomes invisible but I have no idea how to fix that...

GeckoEidechse:
Manually specified progress bar size via CSS now in 4449a4d3592d76141953f90fe6fd80e0e1160d4d

Feels a bit like a hack, probably also is one, but it works and tbh we really need to get a version with progress bar out cause it's the only feature that Viper has that we still don't have ^^

src-vue/src/components/PlayButton.vue
```diff
@@ -85,9 +87,64 @@ export default defineComponent({
...
+                return this.$t("generic.extracting");
+            }
+            return "";
+        },
+        async launchGame() {
+            let unlistenProgress = await appWindow.listen(
+                'northstar-install-download-progress',
+                ({ event, payload }) => {
+                    this.install_or_update = true;
+                    let progress = payload as InstallProgress; // This is bad but don't know how to do it properly
```
Alystrasz, line 126: Why is this "bad"?
If you're talking about the typing, you can type the `listen` generic method invocation with expected answer type:
```typescript
appWindow.listen<InstallType>( ... )
```

src-vue/src/components/PlayButton.vue
```diff
@@ -85,9 +87,64 @@ export default defineComponent({
...
+                    if (progress.state == "DOWNLOADING") {
+                        this.percentage = ((Number(progress.current_downloaded) / Number(progress.total_size)) * 100);
+                        this.color = '#409EFF';
+                        this.status = progress.state;
```
Alystrasz, line 130: This should be outside `if` statements, since it's called by all of them.

src-vue/src/components/PlayButton.vue
```diff
@@ -114,10 +171,24 @@ export default defineComponent({
             />
         </el-option-group>
     </el-select>
+    <el-progress
+        v-if="install_or_update"
```
Alystrasz, line 175: I don't think we should use a `v-if` here, because it removes the progress bar from interface when there's no downloading/extraction process running, meaning the play button will move on progress bar (dis)appearance.
You should use the `v-show` property instead.

src-vue/src/components/PlayButton.vue
```diff
@@ -114,10 +171,24 @@ export default defineComponent({
...
+        :duration="1"
+    >
+    </el-progress>
 </template>
 
 <style scoped>
 
+/* Set progress bar width */
+.el-progress {
+  width: 40vw;
```
Alystrasz, line 189: `vw` unit is related to your window's width, *i.e.* your progress bar width will increase if you change FlightCore's dimensions.

src-vue/src/views/RepairView.vue
```diff
@@ -87,6 +88,16 @@ export default defineComponent({
             });
 
             let install_northstar_result = invoke("install_northstar_caller", { gamePath: game_install.game_path, northstarPackageName: ReleaseCanal.RELEASE });
+
+            const unlistenProgress = await appWindow.listen(
```
Alystrasz, line 92: What is this listener used for?

src-vue/src/components/PlayButton.vue
```diff
@@ -85,9 +87,64 @@ export default defineComponent({
...
+                const status = this.$t("generic.downloading");
+                return `${status}: ${current_downloaded_string}/${total_size_string}`;
+            }
+            if (this.status == "EXTRACTING") {
+                return this.$t("generic.extracting");
+            }
+            return "";
+        },
+        async launchGame() {
+            let unlistenProgress = await appWindow.listen(
```
Alystrasz, line 122: You don't need to keep a reference to this listener if you don't use it later.

Alystrasz, line 122: Actually, why is the listener in the `launchGame` method? This method is triggered when clicking the play button, I don't expect it to setup download progress listener.
This should be done in the `mounted` callback.

Alystrasz (changes requested):
image

Total download size seems incorrect.

src-tauri/src/lib.rs
```diff
@@ -193,7 +215,42 @@ async fn do_install(nmod: &thermite::model::ModVersion, game_path: &std::path::P
...
-    let nfile = thermite::core::manage::download_file(&nmod.url, download_path).unwrap();
+    let last_emit = RefCell::new(Instant::now()); // Keep track of the last time a signal was emitted
+    let nfile = thermite::core::manage::download_file_with_progress(
+        &nmod.url,
+        download_path,
+        |delta, current, total| {
+            if delta != 0 {
+                // Only emit a signal once every 250ms
+                let time_since_last_emit = Instant::now().duration_since(*last_emit.borrow());
+                if time_since_last_emit >= Duration::from_millis(250) {
```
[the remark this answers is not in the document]
GeckoEidechse: When I initially wrote it, it just printed out the progress to cli. When printing every update, I'd would slow down the download from 2sec on my 1gbps connection to 30-40sec as it would stall the download for every print.

I'm assuming pushing out an event everytime has similar performance impact. With Northstar being ~70MB it's 70k events when in reality we don't need to push out every single one but just need to update in a timely basis to give the impression of fluent progress.

Could reduce the latency between updates to e.g. 100ms if we want.

Ironically the faster the update rate, the slower the download speed will be xD

src-vue/src/components/PlayButton.vue
GeckoEidechse: Yeah, this was originally intentional but now did I realise that the rest of the UI doesn't change in size on resizing it's actually kinda stupid to do it like that ^^"

src-vue/src/views/RepairView.vue
GeckoEidechse: It's never used, I just copy and pasted code together until it worked ^^"

src-tauri/src/lib.rs
```diff
@@ -193,7 +215,42 @@ async fn do_install(nmod: &thermite::model::ModVersion, game_path: &std::path::P
...
+                        )
+                        .unwrap();
+                    *last_emit.borrow_mut() = Instant::now();
+                }
+            }
+        },
+    )
+    .unwrap();
+
+    window
```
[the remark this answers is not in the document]
GeckoEidechse: I know but I have to figure out how to get the information in the backend first which is gonna take me a bit.

Given that extraction time is rather static compared to download time I focused on finishing up what I already have. Probably best to add that in a separate PR so that we can roll out download progress bar ASAP ^^

src-tauri/src/lib.rs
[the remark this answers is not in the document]
Alystrasz: I highly suspect the `console.log` invocations slowing down the UI (I don't see how download speed would be impacted, since actual downloading is done by the backend?).
Does thermite send an event for each downloaded kilobyte?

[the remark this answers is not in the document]
GeckoEidechse: > Does thermite send an event for each downloaded kilobyte?

`libthermite` does a callback for every KiB, yes. From there I can decide whether I actually want to fire an event or not ^^

src-vue/src/components/PlayButton.vue
Alystrasz: Addressed in 9982d2e35a201e9481c2ef9f3432d6a52adad969

Alystrasz: Addressed in 9982d2e35a201e9481c2ef9f3432d6a52adad969

src-vue/src/components/PlayButton.vue
Alystrasz: Addressed in 9bb4eef1c175dbb07af1fbab283dbdd064914e7f

src-vue/src/components/PlayButton.vue
Alystrasz: Addressed in 0ef0c49de0e0fdd9dd6e4a0d36b6892ccb86b1be

GeckoEidechse:
> @GeckoEidechse I reworked your PR a bit, feel free to review again (I can't request your review through GitHub since you're the PR author).

Looks good so far, many thanks <3

I'll try to address the remaining issues now. Maybe I'll come across anything in the process :P

src-vue/src/components/PlayButton.vue
GeckoEidechse: Done in 834ab9b7ce66e3c678886ef0a12a96de484e88d0

src-vue/src/components/PlayButton.vue
GeckoEidechse: Technically `InstallProgress` but yes. Thing is, when I try that I get:

> `'InstallProgress' only refers to a type, but is being used as a value here.ts(2693)`

GeckoEidechse: Nevermind, I had typo in the syntax. Done properly in 48e7a2d25934f7e2c7873b3a1a291455f13475e0 now ^^

src-vue/src/views/RepairView.vue
GeckoEidechse: Addressed in a9180cfbecdd3affc860e52af2b43f740cfee371

GeckoEidechse:
Alright, done again from my side now. Back to you for review :P

PR #203, opened 2023-03-06
Add buttons to download PRs

### LINKED ISSUE #201: Add additional button to download PR files
Add an additional button to download CI builds. Sometimes one has a specialised setup or doesn't want to overwrite the existing install (as is the case for Launcher PRs) so it would be useful to have an additional button to download the PR, given that we already have all the code for finding the newest build.

GeckoEidechse:
Adds buttons to just download PR instead of installing them.
In both cases (Launcher and Mods) the button simply opens the download link in the user's default browser.

Idea for feature was conceived in discussion with @EladNLG where they mentioned that grabbing launcher download link is cumbersome to do.

Rendered:

image

Closes #201

### DIFF
--- a/src-vue/src/components/PullRequestsSelector.vue
+++ b/src-vue/src/components/PullRequestsSelector.vue
@@ -56,8 +56,6 @@
 import { defineComponent } from 'vue'
 import { PullRequestType } from '../../../src-tauri/bindings/PullRequestType';
 import { PullsApiResponseElement } from '../../../src-tauri/bindings/PullsApiResponseElement';
-import { invoke } from "@tauri-apps/api";
-import { ElNotification } from "element-plus";
 
 export default defineComponent({
     name: 'PullRequestsSelector',
--- a/src-vue/src/components/PullRequestsSelector.vue
+++ b/src-vue/src/components/PullRequestsSelector.vue
@@ -3,18 +3,12 @@
         <el-collapse @change="onChange">
             <el-collapse-item title="Launcher PRs" name="1">
                 <p v-if="pull_requests_launcher.length === 0">
-                    <el-progress
-                        :show-text="false"
-                        :percentage="100"
-                        status="warning"
-                        :indeterminate="true"
-                        :duration="1"
-                        style="margin: 15px"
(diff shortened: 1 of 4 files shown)


Alystrasz:
I don't get why'd you add a download button while we have an install feature, can you enlighten me about this?

GeckoEidechse:
> I don't get why'd you add a download button while we have an install feature, can you enlighten me about this?

It appeared out of a discussion on Discord. There are occasions where one just wants to download the files to inspect them or send them to someone else etc.

Also we still don't properly support profiles in FlightCore, so giving a way to just download the files quickly and install them manually is helpful.

Some contributors are also suspicious of anything that installs things automatically so this helps them to work around that "issue". ^^

src-vue/src/components/PullRequestsSelector.vue
```diff
@@ -30,17 +24,15 @@
...
-                        :indeterminate="true"
-                        :duration="1"
-                        style="margin: 15px"
-                    />
+                    <el-progress :show-text="false" :percentage="100" status="warning" :indeterminate="true" :duration="1"
+                        style="margin: 15px" />
                 </p>
                 <el-card v-else shadow="hover" v-for="pull_request in pull_requests_mods" v-bind:key="pull_request.url">
                     <el-button type="primary" @click="installModsPR(pull_request)">Install</el-button>
+                    <a target="_blank"
```
Alystrasz, line 32: Why do you put this button inside a link tag, whereas you only use a button to install launcher PRs?
This prevents applying the same style to both buttons.

Alystrasz (changes requested):

src-vue/src/components/PullRequestsSelector.vue
GeckoEidechse: Cause that was the quickest way I found to implement the feature... I guess I should instead have the button call some TypeScript that opens the browser?

Alystrasz: Yeah, just like you did with `downloadLauncherPR`

PR #206, opened 2023-03-10
feat: Add some breadcrumbs printing install path

GeckoEidechse:
This will show up in sentry dashboard on crash and will help me solve a thread crash that's caused by Northstar failing to get installed.

image


Alystrasz (approved):
Confirmed running and building on Windows.

PR #207, opened 2023-03-12
refactor: Move buttons to repair window

GeckoEidechse:
Moves buttons previously located to the new repair window that was added in #129 

image


Alystrasz (approved):
Nice copy/pasting skills :)
Confirmed working on Windows.

PR #208, opened 2023-03-12
feat: Move open repair window button to settings

GeckoEidechse:
Moving the button from DeveloperView to Settings means that it's now accessible to the end-user, essentially "releasing" that feature ^^

image


Alystrasz (approved):
A formality, really.

PR #210, opened 2023-03-14
build: Replace unmaintained action

### LINKED ISSUE #189: Update used Actions in CI

GeckoEidechse:
Replaces all actions of [`actions-rs`](https://github.com/actions-rs) with ones by other authors that are actually maintained.

TODO:
- [x] Replace all actions on push
- [x] Replace all actions on release

When done, nearly closes #189

### DIFF
--- a/.github/workflows/push-test.yml
+++ b/.github/workflows/push-test.yml
@@ -73,8 +73,7 @@ jobs:
       uses: actions/setup-node@v3
       with:
         node-version: 16
-    - name: install Rust stable
-      uses: actions-rs/toolchain@v1
+    - uses: dtolnay/rust-toolchain@stable
       with:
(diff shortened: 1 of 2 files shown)


GeckoEidechse:
Given that it builds it should already be fine cause if I messed something up with the actions, it wouldn't  even build in the first place.

Alystrasz (approved):
As previously said, CI executes successfully, so new actions seems to fit.
I confirm new build works on Windows.

PR #211, opened 2023-03-14
fix: Add error handling if default EA App path

### LINKED ISSUE #171: Detect default EA install location and give warning
"C:\Program Files\EA<something>" cannot be written to by other processes due to some Windows directory permission stuff.

To resolve, the solution is to move the game install to a different location.

We should detect such an issue and guide the user to the right location to resolve.

GeckoEidechse:
We cannot install to `C:\Program Files\EA Games\Titanfall2` cause Windows™
As such if install fails we check for that path and show according error message.

I would kinda also wanna add a link to wiki but AFAIK links won't render clickable so it's not really super helpful...

Closes #171

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -249,7 +249,8 @@ pub async fn install_northstar(
         Err(err) => {
             if game_path
                 .to_lowercase()
-                .contains(&r#"C:\Program Files\"#.to_lowercase()) // default is `C:\Program Files\EA Games\Titanfall2`
+                .contains(&r#"C:\Program Files\"#.to_lowercase())
+            // default is `C:\Program Files\EA Games\Titanfall2`
             {
                 return Err(
                     "Cannot install to default EA App install path, please move Titanfall2 to a different install location.".to_string(),
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -239,12 +239,26 @@ pub async fn install_northstar(
         ..Default::default()
     });
 
-    do_install(
+    match do_install(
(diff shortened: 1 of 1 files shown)


AnActualEmerald:
A plain html link probably won't work but you can open links from within a tauri app https://tauri.app/v1/api/js/shell#open

GeckoEidechse:
> > AFAIK links won't render clickable
> 
> WDYM? We have clickable links on Thunderstore mods cards, for instance.

Right now it's giving a system popup (which arguably is justified cause Northstar failing to install is kinda a big deal). But even if changed I'll still need to essentially stick the HTML into the returned string to show as a link.

At some point I need to refactor error returns to an enumerator so that we can localisation and all that but I first need to learn how to do that which will take me at least a week with everything else going on ^^"

Anyway, I'm happy to take any suggestions to improve the state in the short term :P

src-tauri/src/lib.rs
```diff
@@ -239,12 +239,26 @@ pub async fn install_northstar(
...
         std::path::Path::new(game_path),
     )
     .await
-    .unwrap();
+    {
+        Ok(_) => (),
+        Err(err) => {
+            if game_path
+                .to_lowercase()
+                .contains(&r#"C:\Program Files\EA Games\Titanfall2"#.to_lowercase())
```
Alystrasz, line 252: Can the EA App games path be different from this one?

Alystrasz: Why not checking for `C:\Program Files\` only then?

GeckoEidechse: Fair enough, changed in 8992810e0fcce01ba9b7b0c480db8d80009e1f24

PR #217, opened 2023-03-18
refactor: Fix PR downloads permission issue

### LINKED ISSUE #212: PR installing broken on installed release builds
So turns out that for some reason installing a PR on installed release builds actually fails with a panic.

Reason is lack of write permissions.

Need to double check but IIRC we download to current dir which might not be accessible? Maybe we should just download to Titanfall2 dir instead?

GeckoEidechse:
Avoids any folder write permission issues by writing into memory and/or temp dir in Titanfall2 folder.

This started out as a small bug fix PR but uncovered a bunch of issues that prompted me to refactor most of the code entirely.

TODO:

- [x] Download Launcher PRs into Titanfall2 dir
- [x] Download Mods PRs into Titanfall2 dir
- [x] Reverify above two
- [x] Compare old and new download code
- [x] Sanity checks

Closes #212


Alystrasz (approved):
Code looks good, feature works as expected on Windows with CI build.

PR #221, opened 2023-03-19
feat: Increase time mods PR install notification is shown

GeckoEidechse:
So as to not destroy it while user is still reading it. Changed from the default `4.5s` to `7s`.

1 line change 👀


Alystrasz (approved):
Confirmed working on Windows.

PR #224, opened 2023-03-20
fix: Handle error when Northstar download fails

GeckoEidechse:
In case of download failure return an error message.

Should resolve https://northstar-kv.sentry.io/issues/4014183248/

image

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -195,7 +195,10 @@ async fn do_install(nmod: &thermite::model::ModVersion, game_path: &std::path::P
     let download_path = format!("{}/{}", download_directory.clone(), filename);
     println!("{}", download_path);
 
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Sentry doesn't let me link Sentry<->GitHub issues without a subscription which costs 20$+ a month >:(

Alystrasz:
What is the expected behaviour of download failure?
I tried to provoke a crash shutting down my Internet connection during a forced reinstall, but only got a "Unable to open zip archive" error message.

GeckoEidechse:
> What is the expected behaviour of download failure?

It should be `Failed downloading Northstar <error message here>`

> I tried to provoke a crash shutting down my Internet connection during a forced reinstall, but only got an "Unable to open zip archive" error message, both with `main` and this branch.

Did it maybe download too fast so by the time you disconnected the connection the download already finished? Although that wouldn't explain the error regarding opening the zip file...

PR #182, opened 2023-02-14
feat: i18n

Alystrasz:
Adds frontend translation.

Closes #172.

---

##### TODOs

- [x] Translate interface
- [x] Add component to select language
- [ ] Translate backend messages (mainly errors)
- [ ] Study automatic translation of natively-English content (mods' descriptions for instance)
- [ ] Weblate integration?

PR #227, opened 2023-03-21
fix: Address regression around persistent store

### LINKED ISSUE #233: Storing persistence is broken
Tested on Linux rn and broken since at least `v1.10.1` of FlightCore.

Probably #176 that broke it.

GeckoEidechse:
As discussed in #233 saving persistent store was broken some time ago. Calling `save` explicitly  seems to resolve it.

For good measure, also bumped plugin-store version and updated pulled dependency as [described in their README](https://github.com/tauri-apps/plugins-workspace/tree/dev/plugins/store).

- Closes #233 
- Unblocks #182

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -43,7 +43,7 @@ sysinfo = "0.26.2"
 # HTTP requests
 reqwest = { version = "0.11", features = ["blocking"] }
 # Persistent store for settings
-tauri-plugin-store = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "dev" }
+tauri-plugin-store = { git = "https://github.com/tauri-apps/plugins-workspace", rev = "5a6abd3203dc94c38f96d0c4bf7ecbef399f8c25" }
 # JSON5 parsing support (allows comments in JSON)
 json5 = "0.4.1"
 # Async recursion for recursive mod install
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -43,7 +43,7 @@ sysinfo = "0.26.2"
 # HTTP requests
 reqwest = { version = "0.11", features = ["blocking"] }
 # Persistent store for settings
-tauri-plugin-store = { git = "https://github.com/tauri-apps/tauri-plugin-store", rev = "9bd993aa67766596638bbfd91e79a1bf8f632014" }
+tauri-plugin-store = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "dev" }
 # JSON5 parsing support (allows comments in JSON)
 json5 = "0.4.1"
 # Async recursion for recursive mod install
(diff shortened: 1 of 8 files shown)


GeckoEidechse:
Need to do some testing to confirm still.

GeckoEidechse:
Yeah that doesn't seem to fix #233

GeckoEidechse:
Even trying to reduce tauri API version again does not resolve the issue T-T

GeckoEidechse:
Still not fixed btw

GeckoEidechse:
Fixed with 6b087b1448d3e86a04359a752e2cc6ae3b886cdd on Linux. Cannot test Windows atm.

src-vue/src/views/SettingsView.vue
```diff
@@ -78,9 +78,10 @@ export default defineComponent({
             get(): boolean {
                 return this.$store.state.enableReleasesSwitch;
             },
-            set(value: boolean): void {
+            async set(value: boolean): Promise<void> {
```
GeckoEidechse, line 81: Uh, the `Promise` return value was introduced by running a quick fix action in vscode. Not sure if correct ^^"

Alystrasz:
So basic issue was not saving store after inserting new keys?
I don't understand how persistence worked somehow while we didn't call `save`?

GeckoEidechse:
> So basic issue was not saving store after inserting new keys? I don't understand how persistence worked somehow while we didn't call `save`?

So basically if you don't call `save()` it only saves persistence when exiting the application which was how we did saving so far. For some reason this broke at some point for reason unknown to me. Calling `save()` explicitly seems to be enough to "fix" it.
A proper fix would require investigating the issue upstream which is something I don't have the time/knowledge to do ^^"

src-vue/src/views/SettingsView.vue
Alystrasz: Yeah it's correct

src-tauri/Cargo.toml
```diff
@@ -43,7 +43,7 @@ sysinfo = "0.26.2"
 # HTTP requests
 reqwest = { version = "0.11", features = ["blocking"] }
 # Persistent store for settings
-tauri-plugin-store = { git = "https://github.com/tauri-apps/tauri-plugin-store", rev = "9bd993aa67766596638bbfd91e79a1bf8f632014" }
+tauri-plugin-store = { git = "https://github.com/tauri-apps/plugins-workspace", branch = "dev" }
```
Alystrasz, line 46: Could you point to a commit instead of a branch?

Alystrasz (commented):
Apart from the comment, looks good; confirmed working on Windows.

src-tauri/Cargo.toml
GeckoEidechse: Shouldn't really be necessary but given that the plugin store is still under active development it might be for the best.

Resolved in 22ac2cafd7fb5946a5ba99044572fb3316e59761

Alystrasz (approved):

Issue #225, opened 2023-03-20
Run `cargo clippy` over the project and fix any issues it identifies


GeckoEidechse:
[`clippy`](https://doc.rust-lang.org/stable/clippy/index.html) is a tool to autodetect common programming mistakes in Rust code and point them out.

We should run it over the project and address any changes it might suggest.

In the future we should also add it to CI but I first need to improve our CI config so as to not increase runtime more than it is already ^^"

GeckoEidechse:
With #255 the only still open warning generated by `cargo clippy` is:

```
warning: this call to `as_ref` does nothing
  --> src\thunderstore\mod.rs:69:72
   |
69 |     let to_remove_set: HashSet<&str> = BLACKLISTED_MODS.iter().map(|s| s.as_ref()).collect();
   |                                                                        ^^^^^^^^^^ help: try this: `s`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_asref
   = note: `#[warn(clippy::useless_asref)]` on by default

warning: `app` (bin "app") generated 1 warning
```

However changing the line as suggested by clippy introduces new issues. Would need your help to resolve that one @catornot ^^

catornot:
I am unfortunately ~2000 kilometers from my computer so I can't really tell you the best way of fixing it. If I thought clippy warnings were dump I would just allow them specifically for unsafe docs warning which is dumb imo for internal stuff. Anyway if that warning is hard to fix you could just allow it for the specific line by adding `#[allow(clippy::warning)]`

catornot:
> However changing the line as suggested by clippy introduces new issues. Would need your help to resolve that one @catornot ^^

ok this might work 

```rust
let to_remove_set: HashSet<&str> = BLACKLISTED_MODS.iter().copied().collect();
```

GeckoEidechse:
> ok this might work
> 
> ```rust
> let to_remove_set: HashSet<&str> = BLACKLISTED_MODS.iter().copied().collect();
> ```

Yup, that looks like it did the trick. Got a PR for it now in #261 ^^

PR #229, opened 2023-03-22
fix: Address clippy warnings regarding format!()

GeckoEidechse:
Use the variable name directly inside the string.

Partially addresses #225


Alystrasz (approved):
Only style changed here.
Confirmed working on Windows.

PR #231, opened 2023-03-23
fix: Remove unnecessary borrows

GeckoEidechse:
According to `clippy`

Partially addresses #225


Alystrasz (approved):
Works fine on Windows.

PR #232, opened 2023-03-23
fix: Adjust code for single iteration loop

GeckoEidechse:
`clippy` said this is way is better...

Partially addresses #225

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -339,13 +339,10 @@ pub fn check_origin_running() -> bool {
 
 /// Checks if Northstar process is running
 pub fn check_northstar_running() -> bool {
-    let s = sysinfo::System::new_all();
-    if let Some(_process) = s.processes_by_name("NorthstarLauncher.exe").next() {
-        // check here if this is your process
-        // dbg!(process);
-        return true;
-    }
-    false
+    sysinfo::System::new_all()
+        .processes_by_name("NorthstarLauncher.exe")
+        .next()
+        .is_some()
 }
 
 /// Helps with converting release candidate numbers which are different on Thunderstore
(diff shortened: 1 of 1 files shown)


src-tauri/src/lib.rs
```diff
@@ -350,7 +350,7 @@ pub fn check_origin_running() -> bool {
 /// Checks if Northstar process is running
 pub fn check_northstar_running() -> bool {
```
catornot, line 351: suggestions doesn't work so here is another one liner :trollface: 
```rust
pub fn check_northstar_running() -> bool {
    sysinfo::System::new_all().processes_by_name("NorthstarLauncher.exe")
}
```

catornot (commented):
do you like one liners?

src-tauri/src/lib.rs
```diff
@@ -333,14 +333,14 @@ pub fn launch_northstar(
...
         // dbg!(process);
         // There's at least one Origin process, so we can launch
         return true;
     }
     // Alternatively, check for EA Desktop
-    for _process in s.processes_by_name("EADesktop.exe") {
+    if let Some(_process) = s.processes_by_name("EADesktop.exe").next() {
         // There's at least one EADesktop process, so we can launch
         return true;
     }
```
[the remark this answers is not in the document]
GeckoEidechse: Implemented in ea39434f424cceb3757a089a4f836c17166cd954

src-tauri/src/lib.rs
GeckoEidechse: Implemented in 4c5fc48c41370ad17aef1bbad2fd7313ee49808a

Alystrasz (approved):
Confirmed working on Windows.

PR #223, opened 2023-03-19
feat: Proper logging

GeckoEidechse:
Uses `log` and `sentry-log` to do proper logging

TODO:
- [ ] Check if logging initialised correctly
- [x] Replace existing breadcrumbs with logger statements (slimmer and more readable
- [ ] Replace any `println!()`s with logging statements
- [ ] Add timestamps to logging

Closes #219

Issue #219, opened 2023-03-19
Use proper logging library for backend instead of occasional `println!()`


GeckoEidechse:
Having a proper logging library instead of the occasional `println()!` would allow us to do proper logging with filters (`warn`, `info`, `debug`, ...) etc.

Just need to figure out what the defacto standard library for this is ^^

PR #236, opened 2023-03-24
refactor: Replace all `println!()` with log statements

GeckoEidechse:
Replace more `println!()` with calls to `log::info`, `log::warn`, etc

Continuation of
- #219
- #223


catornot:
why so many commits?

Alystrasz (approved):
refactor: Replace all println with calls to logger

GeckoEidechse:
> why so many commits?

I prefer doing smaller incremental changes and then validating those, so in this case replacing all calls in a function and then testing it to verify.
While a bit more tedious, it did help me find a bunch of smaller issues in the process ^^

Also, commit stats go brrrrr >:)

catornot:
I just push to main

Issue #238, opened 2023-03-25
Instructions for building.


begin-theadventure:
Would it be possible to add instructions for building from source code? ~~I tried, but failed (I can add more information if needed).~~ Never mind, I managed to build it by looking at the [workflow file](https://github.com/R2NorthstarTools/FlightCore/actions/runs/4462846117/workflow) although it also bundles deb and appimage.

GeckoEidechse:
I assumed the [build instructions](https://github.com/R2NorthstarTools/FlightCore/blob/main/docs/DEVELOPMENT.md#setup), which are [linked in the main README](https://github.com/R2NorthstarTools/FlightCore#development) are enough but if not I'm happy to improve them on any point where you might see an issue with them. ^^

begin-theadventure:
They can be helpful, although they won't build, unlike those in the workflow file.
I tried to build in src-tauri, but it does the same thing as in the main folder, bundles the AppImage and deb, so I would like to know how to build just the binary. Maybe I'd have to change some configuration to do that?
The commands I've used: `npm install && cd src-vue && npm install && cd ../src-tauri && npm run tauri build` (and `cargo-tauri build`, but it's the same, I guess).
I wonder if such arguments would be a good idea:
`npm run tauri build flight-core` - binary only; appimage/deb if binary exists.

begin-theadventure:
They can be helpful, although they won't build, unlike those in the workflow file.
I tried to build in src-tauri, but it does the same thing as in the main folder, bundles the AppImage and deb, so I would like to know how to build just the binary. Maybe I'd have to change some configuration to do that?
The commands I've used: `npm install && cd src-vue && npm install && cd ../src-tauri && npm run tauri build` (and `cargo-tauri build`, but it's the same, I guess).

GeckoEidechse:
What's the error message you're seeing?

Keep in mind that by default the auto-updater is enabled and it requires a private key to sign the finished build. You can disable the updater by setting the [corresponding value](https://github.com/R2NorthstarTools/FlightCore/blob/c102fdcf8ef00e3c5c46fb9ded37e09b7e92e483/src-tauri/tauri.conf.json#L54) to `false` in `tauri.conf.json`.

As for building "just the binary", you'll have to take that issue upstream. Tauri build scripts for Linux only produce an AppImage and `.deb` package (in the future also Flatpak as far as I can tell).

begin-theadventure:
[...]

PR #240, opened 2023-03-26
docs: Expand on the Setup (build) instructions

begin-theadventure:
#238

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -34,7 +34,7 @@ npx tauri dev
 
 Automatic recompiling on save is enabled for both the Rust and the Typescript/Vue code.
 
-If you want to build, do all of the above (except tauri dev) and:
+If you want to build FlightCore from source code, do the first two commands and:
 
 ```sh
 cd ../src-tauri
@@ -44,21 +44,23 @@ cd ../src-tauri
 npm run tauri build
 ```
 
-This will build the executable and bundles, such as AppImage or .deb.
+This will build the executable and bundles, such as `AppImage`, `.deb` or `.msi`.
 
-To build just the executable, edit [tauri.conf.json](https://github.com/R2NorthstarTools/FlightCore/blob/c102fdcf8ef00e3c5c46fb9ded37e09b7e92e483/src-tauri/tauri.conf.json#L18) in the same folder.
+To build just the executable, edit [tauri.conf.json](https://github.com/R2NorthstarTools/FlightCore/blob/main/src-tauri/tauri.conf.json) in the same folder:
+
+```
+    "bundle": {
+      "active": true,
+```
 
 Change active from `true` to `false`, and bundles won't be included afterward.
 
-To disable the updater change [this value](https://github.com/R2NorthstarTools/FlightCore/blob/c102fdcf8ef00e3c5c46fb9ded37e09b7e92e483/src-tauri/tauri.conf.json#L54) to false.
+To disable the updater (which requires a private key) change this active value to false:
 
-All of this can be simplified into a one-line command:
-
-```sh
-cd src-tauri; sed -i '18s/.*/      "active": false,/' tauri.conf.json; sed -i '54s/.*/      "active": false,/' tauri.conf.json; cd ..; npm install && cd src-vue; npm install && cd ../src-tauri; npm run tauri build
 ```
-
-If you want either of these options, remove the sed commands, 18 is for the former, and 54 is for the latter.
+    "updater": {
+      "active": true,
+```
 
 ## Tauri
 
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -14,7 +14,6 @@ As for splitting logic between _frontend_ and _backend_, state and UI related lo
 
 Make sure you have the necessary dependencies for Tauri installed as described in this link: https://tauri.app/v1/guides/getting-started/prerequisites
 
-
 Then, install `npm` dependencies with
 
 ```sh
@@ -34,6 +33,32 @@ npx tauri dev
 ```
 
 Automatic recompiling on save is enabled for both the Rust and the Typescript/Vue code.
+
+If you want to build, do all of the above (except tauri dev) and:
+
+```sh
+cd ../src-tauri
+```
+
+```sh
+npm run tauri build
+```
+
+This will build the executable and bundles, such as AppImage or .deb.
+
+To build just the executable, edit [tauri.conf.json](https://github.com/R2NorthstarTools/FlightCore/blob/c102fdcf8ef00e3c5c46fb9ded37e09b7e92e483/src-tauri/tauri.conf.json#L18) in the same folder.
+
+Change active from `true` to `false`, and bundles won't be included afterward.
+
+To disable the updater change [this value](https://github.com/R2NorthstarTools/FlightCore/blob/c102fdcf8ef00e3c5c46fb9ded37e09b7e92e483/src-tauri/tauri.conf.json#L54) to false.
+
+All of this can be simplified into a one-line command:
+
+```sh
+cd src-tauri; sed -i '18s/.*/      "active": false,/' tauri.conf.json; sed -i '54s/.*/      "active": false,/' tauri.conf.json; cd ..; npm install && cd src-vue; npm install && cd ../src-tauri; npm run tauri build
+```
+
+If you want either of these options, remove the sed commands, 18 is for the former, and 54 is for the latter.
 
 ## Tauri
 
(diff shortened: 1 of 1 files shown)


docs/DEVELOPMENT.md
```diff
@@ -35,6 +34,32 @@ npx tauri dev
...
+To build just the executable, edit [tauri.conf.json](https://github.com/R2NorthstarTools/FlightCore/blob/c102fdcf8ef00e3c5c46fb9ded37e09b7e92e483/src-tauri/tauri.conf.json#L18) in the same folder.
+
+Change active from `true` to `false`, and bundles won't be included afterward.
+
+To disable the updater change [this value](https://github.com/R2NorthstarTools/FlightCore/blob/c102fdcf8ef00e3c5c46fb9ded37e09b7e92e483/src-tauri/tauri.conf.json#L54) to false.
+
+All of this can be simplified into a one-line command:
+
+```sh
+cd src-tauri; sed -i '18s/.*/      "active": false,/' tauri.conf.json; sed -i '54s/.*/      "active": false,/' tauri.conf.json; cd ..; npm install && cd src-vue; npm install && cd ../src-tauri; npm run tauri build
```
Alystrasz, line 58: I see that you're using raw line numbers here, does it mean that this documentation becomes invalid if we update `tauri.conf.json`?

begin-theadventure: Yes, and the same for links mentioning it. I'll see if I can change it to find it.

begin-theadventure: Yes, and the same for links mentioning it. I'll see if I can change it to find.

GeckoEidechse: To be fully independent of line numbers I'd just use `jq` to parse/edit the JSON instead. Although I don't really see the need to put the script in here in the first place to be honest ^^

docs/DEVELOPMENT.md
```diff
@@ -35,6 +34,32 @@ npx tauri dev
...
 
+If you want to build, do all of the above (except tauri dev) and:
+
+```sh
+cd ../src-tauri
+```
+
+```sh
+npm run tauri build
+```
```
GeckoEidechse, line 45: I'm a bit confused, are saying one should do this or not do this?

docs/DEVELOPMENT.md
```diff
@@ -35,6 +34,32 @@ npx tauri dev
...
+
+This will build the executable and bundles, such as AppImage or .deb.
```
GeckoEidechse, line 47: Don't forget that on Windows it builds an `.msi`.

docs/DEVELOPMENT.md
```diff
@@ -35,6 +34,32 @@ npx tauri dev
...
+
+To build just the executable, edit [tauri.conf.json](https://github.com/R2NorthstarTools/FlightCore/blob/c102fdcf8ef00e3c5c46fb9ded37e09b7e92e483/src-tauri/tauri.conf.json#L18) in the same folder.
```
GeckoEidechse, line 49: `npx tauri build ` will always produce a bundle (`msi`, `deb`, AppImage), there isn't really a way to get just the executable (well, sorta).

docs/DEVELOPMENT.md
begin-theadventure: Ok, I'll delete.

docs/DEVELOPMENT.md
begin-theadventure: "such as `AppImage`, `.deb` or `.msi`", I'll add it, thanks.

docs/DEVELOPMENT.md
begin-theadventure: Is this better?:
If you want to build, do the first two commands and:

docs/DEVELOPMENT.md
begin-theadventure: So editing the config is not the solution?

docs/DEVELOPMENT.md
GeckoEidechse: I mean say what to build, but yeah, ig

docs/DEVELOPMENT.md
GeckoEidechse: Well, editing the config, allows you to build the bundle without having to give a private key (at the cost of not having an self-update mechanism anymore).

begin-theadventure: 49 line in the commit isn't related to the updater.

begin-theadventure: The 49th line in the commit isn't related to the updater.
I don't know what I have to change here.
I've tested this plenty of times, and [bundle active false](https://github.com/R2NorthstarTools/FlightCore/blob/c102fdcf8ef00e3c5c46fb9ded37e09b7e92e483/src-tauri/tauri.conf.json#L18) always disabled the bundle for me.
The only change I can think of is to mention it this way:
```
    "bundle": {
      "active": true,
```
in case of an update to tauri.conf.json.

GeckoEidechse: Oh, never tried changing that xD

Never looked at what would happen when disabling that ^^

docs/DEVELOPMENT.md
```diff
@@ -35,6 +34,34 @@ npx tauri dev
...
 Automatic recompiling on save is enabled for both the Rust and the Typescript/Vue code.
 
+If you want to build FlightCore from source code, do the first two commands and:
+
+```sh
+cd ../src-tauri
+```
+
+```sh
+npm run tauri build
```
GeckoEidechse, line 44: I would shorten it to just:

```suggestion
If you want to build FlightCore from source, run

```sh
npx tauri build
```

begin-theadventure: Will this also take care of npm and vue dependencies?

GeckoEidechse: No. The way I read it it's a continuation of the section above, so like run `npx tauri dev` for dev setup, `npx tauri build` for release build.

begin-theadventure: Doesn't `cd src-vue && npm install` imply that build should be run in src-vue, instead of the main folder?

GeckoEidechse: Kinda but then it needs to be fixed in the section above cause you need to do the same things for `npx tauri dev` and `npx tauri build`.

Given slightly altered instructions for one but not for the other just makes the whole thing more confusing IMO.

begin-theadventure: cd .. && npx tauri dev

GeckoEidechse: > cd .. && npx tauri dev or [cd src-vue && npm install && cd ..](https://github.com/R2NorthstarTools/FlightCore/blob/main/.github/workflows/release.yml#L51)

Essentially it's 
```
npm install
cd src-vue
npm install
cd ..
```
to install node dependencies and then one can run `npx tauri dev` or `npx tauri build`

&nbsp;

>  By the way, will you also change to npx in the workflow file?

Eh idc, `npx` is just `npm run` AFAIK so doesn't really matter for scripting etc. For instructions however it should be consistent though. So either always `npx` or always `npm run`

docs/DEVELOPMENT.md
```diff
@@ -35,6 +34,30 @@ npx tauri dev
...
+To build just the executable, edit [tauri.conf.json](https://github.com/R2NorthstarTools/FlightCore/blob/main/src-tauri/tauri.conf.json) in the same folder:
+
+```json
+    "bundle": {
+      "active": true,
+```
+
+Change active from `true` to `false`, and bundles won't be included afterward.
+
+To disable the updater (which requires a private key) change this active value to false:
```
GeckoEidechse, line 54: Better formatting:

```suggestion
To disable the updater (which requires a private key) change this active value to `false`:
```

docs/DEVELOPMENT.md
```diff
@@ -35,6 +34,30 @@ npx tauri dev
...
+This will build the executable and bundles, such as `AppImage`, `.deb` or `.msi`.
+
+To build just the executable, edit [tauri.conf.json](https://github.com/R2NorthstarTools/FlightCore/blob/main/src-tauri/tauri.conf.json) in the same folder:
+
+```json
+    "bundle": {
+      "active": true,
+```
+
+Change active from `true` to `false`, and bundles won't be included afterward.
```
GeckoEidechse, line 52:
```suggestion
Change `active` from `true` to `false`, and bundles won't be included afterwards.
```

begin-theadventure: Afterward/afterwards are both correct, just more used depending on the country, USA/UK etc.

GeckoEidechse (approved):
Should be good now ^^

begin-theadventure:
I noticed this after the merge:
To build just the executable, edit [tauri.conf.json](https://github.com/R2NorthstarTools/FlightCore/blob/main/src-tauri/tauri.conf.json) in the ~~same~~ src-tauri folder.

GeckoEidechse:
eh, just make a new PR fixing it ig ¯\\\_(ツ)\_/¯

PR #244, opened 2023-03-26
fix: Replace `<=0` with `.is_empty()`

GeckoEidechse:
Based on `clippy` output.

Partially addresses #225


GeckoEidechse:
Draft cause I still need to test it.

PR #246, opened 2023-03-27
fix: Another set of various clippy fixes

GeckoEidechse:
Various fixes based on `clippy` output.

With this PR we are down to 3 more warnings, 2 of which will require me to also change something in front-end, 1 I will need some assistance to resolve.

Partially addresses #225


GeckoEidechse (commented):
Tested working on Windows

PR #247, opened 2023-03-28
refactor: Style

Alystrasz:
Some little visual changes:
* Remove blur from title bar
* ~~Remove hashed title bar style when dev mode is disabled~~
* Unify settings panels style

Capture d’écran de 2023-03-28 17-33-29


src-vue/src/plugins/store.ts
```diff
@@ -94,6 +95,11 @@ export const store = createStore<FlightCoreStore>({
             if (!state.developer_mode) {
                 store.commit('updateCurrentTab', Tabs.PLAY);
             }
+
+            let menu_bar_handle = document.querySelector('#fc_menu-bar');
+            if (menu_bar_handle !== null) {
+                menu_bar_handle.classList.toggle('developer_build');
+            }
```
GeckoEidechse, line 102: Toggling developer mode shouldn't change the top bar stripes IMO. The stripes indicate _debug build_ not that developer mode was activated.

The reason for this being that when I have local debug build running that I'm developing on rn as well as a release build, I want to be able to detect, which is which.

For example, one of these is a downloaded PR arctifact, the other one is locally compiled

image

do you know which one is which? ;)

Alystrasz: Oh I didn't know this style represented build state and not dev mode state.
Addressed in 9c5bfe23cc890ead0e94dab60edb70824fd80890

src-vue/src/views/mods/LocalModsView.vue
```diff
@@ -1,33 +1,35 @@
...
+                    <el-button type="danger">
+                        {{ $t('mods.local.delete') }}
+                    </el-button>
+                </template>
+            </el-popconfirm>
+            {{ mod.name }}
+            <span v-if="mod.version != null">(v{{ mod.version }})</span>
+            <img
+                v-if="mod.thunderstore_mod_string != null"
+                title="This Northstar mod is part of a Thunderstore mod"
```
GeckoEidechse, line 27: Accidental i18n regression

src-vue/src/views/SettingsView.vue
```diff
@@ -40,25 +41,30 @@
...
+                <div class="fc_parameter__panel">
+                    <h3>{{ $t('settings.testing') }}</h3>
+                    <span>
+                        {{ $t('settings.enable_test_channels') }}
+                        <el-switch v-model="enableReleasesSwitch"></el-switch>
+                    </span>
                 </div>
-                <br />
-                <br />
-                UI design inspired by <el-link :underline="false" target="_blank" href="https://github.com/TFORevive/tforevive_launcher/" type="primary">TFORevive Launcher</el-link> (not yet public)
```
GeckoEidechse, line 55: You finally wanna get rid of it? haha ^^

src-vue/src/plugins/store.ts
```diff
@@ -87,13 +88,18 @@ export const store = createStore<FlightCoreStore>({
         checkNorthstarUpdates(state) {
             _get_northstar_version_number(state);
         },
-        toggleDeveloperMode(state) {
+        toggleDeveloperMode(state, affectMenuStyle = false) {
```
GeckoEidechse, line 91: Not sure how I fell about this as it kinda combines developer and debug mode but on the other side it's still the result as before, so idc too much tbh ^^

src-vue/src/views/SettingsView.vue
Alystrasz: Yeah, since this is basically a lie 🤡

src-vue/src/views/mods/LocalModsView.vue
Alystrasz: Addressed in e32dd824d8fd54047a13f8f994a497434853e3be

src-vue/src/plugins/store.ts
Alystrasz: So that when you change your mind, fixing it will be fast 😄

GeckoEidechse (approved):
Found no issues in testing and code looks fine as far as I can tell ^^

PR #248, opened 2023-03-29
docs: Explanation on expanding language selector

GeckoEidechse:
Add explanation on how to add a new language to the language selector component

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -221,6 +221,29 @@ export const i18n = createI18n({
 });
 ```
 
+In order to be able to select it, make sure to that it to the `LanguageSelector` componenent in `LanguageSelector.vue`.
+
+```vue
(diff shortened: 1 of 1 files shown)


docs/DEVELOPMENT.md
```diff
@@ -221,6 +221,29 @@ export const i18n = createI18n({
 });
 ```
 
+In order to be able to select it, make sure to that it to the `LanguageSelector` componenent in `LanguageSelector.vue`.
```
Alystrasz, line 224:
```suggestion
In order to be able to select it, make sure to that it to the `LanguageSelector` componenent in `src-vue/src/components/LanguageSelector.vue`.
```

Alystrasz (changes requested):

Alystrasz (approved):
Good point I completely missed 😄

PR #249, opened 2023-03-29
feat: Initial German translation

GeckoEidechse:
Attempted to add German translation. Should be fine for the most part but would like some other German native speaker to double check ^^

TODO:
- [x] Merge `main` and switch translations to JSON with #253 
- [x] Proofreads


GeckoEidechse:
@NoCatt said he didn't find any more issues after the ones I fixed so I guess proofreading is done ^^

src-vue/src/i18n/lang/de.json
```diff
@@ -0,0 +1,156 @@
...
+
+        "online": {
+            "no_match": "Keine passenden Mods gefunden.",
+            "try_another_search": "Versuche eine andere Suchanfrage!"
+        },
+
+        "menu": {
+            "local": "Lokal",
+            "online": "Online",
+            "filter": "Filter",
```
Alystrasz, line 55: Forgotten translation, or do you use the same word in German? :)

GeckoEidechse: Actually the same word in German, ye ^^

Alystrasz (approved):

PR #253, opened 2023-03-30
refactor: Convert localization files from TypeScript to JSON format

Alystrasz:
Weblate only supports formats like JSON out of the box. As such converting translations to JSON helps with weblate integration.


GeckoEidechse (commented):
Don't forget updating the docs ;)

GeckoEidechse (approved):
Confirmed still working on Linux

PR #255, opened 2023-04-02
fix: Change enum value casing

GeckoEidechse:
Change enum value casing as suggested by `clippy`

Partially addresses #225


src-vue/src/components/PullRequestsSelector.vue
```diff
@@ -71,10 +71,10 @@ export default defineComponent({
         onChange(e: Object) {
             const openedCollapseNames = Object.values(e);
             if (openedCollapseNames.includes('1') && this.pull_requests_launcher.length === 0) {
-                this.getPullRequests('LAUNCHER');
+                this.getPullRequests('Launcher');
```
Alystrasz, line 74: Actually, why is `PullRequestType` not an enumeration?
I'd expect something like this:
```javascript
this.getPullRequests(PullRequestType.Launcher);
```

src-vue/src/components/PullRequestsSelector.vue
```diff
@@ -71,10 +71,10 @@ export default defineComponent({
...
             }
             if (openedCollapseNames.includes('2') && this.pull_requests_mods.length === 0) {
-                this.getPullRequests('MODS');
+                this.getPullRequests('Mods');
```
Alystrasz, line 77: Same remark here

Alystrasz (changes requested):
I might have omitted an issue while merging PR-install feature :)
Otherwise, code looks good and runs fine on Ubuntu.

src-vue/src/components/PullRequestsSelector.vue
GeckoEidechse: Cause for some reason `ts-rs` exports it as a set of strings and I don't have anyway of changing that unfortunately ._.

https://github.com/R2NorthstarTools/FlightCore/pull/255/files#diff-68931ccb7f8858bb1c8ae0d7109db6ab91c0531c6233d5e036a1e5faf4277bebR3

See also https://github.com/Aleph-Alpha/ts-rs/issues/23

Alystrasz (approved):

PR #256, opened 2023-04-05
build: Generate release draft on tag

GeckoEidechse:
Instead of creating a release and then having CI build the files, have CI build files on tag and then upload them to draft release.

Previously when making a release, this release wouldn't have any files ready until build finished (~30min). During this time the download link on the main README would be broken for example.

This PR changes it to upload files to a release draft and then that release draft can be turned into an actual release, meaning that now there's no time where a GitHub release has no files attached.

This could also be extended in the future to auto-generate release notes like how we do currently from within FligthCore dev tools.

Tested successful run in test repo here: 
https://github.com/GeckoEidechse/Temp-FlightCore/actions/runs/4598431296

and resulting draft (only I have access to that in this case cause personal repo):
https://github.com/GeckoEidechse/Temp-FlightCore/releases/tag/untagged-eda3413aa1a20df8ded7

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,11 @@
 name: "release"
 on:
-  release:
-    types: [ created ]
+  push:
+    tags:
+      - 'v*'
+
+permissions:
+  contents: write # Needed to write to GitHub draft release
 
 jobs:
   # Ensure version numbers in various places match up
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
@Alystrasz any remarks from your end or should I just merge? ^^

Alystrasz:
Nah, this looks good, you already tested it, I got nothing to add :)

GeckoEidechse:
Aaand it failed, lol
https://github.com/R2NorthstarTools/FlightCore/actions/runs/4623158983/jobs/8176681763

Will investigate ^^"

GeckoEidechse:
So after pushing the wrong tag like 3 times, I think I finally got the right tag pushed now ^^

https://github.com/R2NorthstarTools/FlightCore/actions/runs/4623250120

PR #257, opened 2023-04-06
build: Use `npm clean-install`

GeckoEidechse:
Use `npm clean-install` instead of `npm install`. `npm clean-install` is recommended over `npm install` for CI pipelines.

https://docs.npmjs.com/cli/v9/commands/npm-ci


Alystrasz (approved):
Confirmed working on Ubuntu.

PR #261, opened 2023-04-10
fix: Remove unnecessary `as_ref` call

GeckoEidechse:
This should remove the last `clippy` warning ^^


GeckoEidechse (commented):
Confirmed working in testing on Linux

PR #265, opened 2023-04-14
feat: Persist dev mode

### LINKED ISSUE #214: Persist dev mode state
The PR tool is only available in dev mode and it's annoying to enable it every time I start flightcore

Alystrasz:
Saves dev mode state in persistent store, so it's automatically enabled on app launch if it was previously enabled.

Closes #214.

### DIFF
--- a/src-vue/src/plugins/store.ts
+++ b/src-vue/src/plugins/store.ts
@@ -88,17 +88,18 @@ export const store = createStore<FlightCoreStore>({
         checkNorthstarUpdates(state) {
             _get_northstar_version_number(state);
         },
-        async toggleDeveloperMode(state, affectMenuStyle = false) {
+        async toggleDebugMode(_state) {
+            let menu_bar_handle = document.querySelector('#fc_menu-bar');
+            if (menu_bar_handle !== null) {
+                menu_bar_handle.classList.toggle('developer_build');
+            }
+        },
+        async toggleDeveloperMode(state) {
             state.developer_mode = !state.developer_mode;
 
             // Reset tab when closing dev mode.
             if (!state.developer_mode) {
                 store.commit('updateCurrentTab', Tabs.PLAY);
-            }
-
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Actually wanted to work on this the other day but then some other stuff happened and I was too busy again, so many thanks for taking over <3

What I was thinking by changing dev mode to be persistent we could actually drop auto-enabling it on debug builds.

The main reason it was auto-enabled on debug builds was cause having to spam click version number to switch to dev view to test some new feature and then having to repeat that process on each backend-end recompile got really annoying. By storing dev mode in persistent store, one can in turn now just enable it once  now and it'll stay on, so auto-enabling it on debug builds is technically no longer needed ^^

Would also help in decoupling debug builds vs dev mode ;)

Thoughts?

Alystrasz:
I agree with you!
Your points were addressed in a0aef64a71a0dd0b4d65735313031d6643ea0987.

src-vue/src/plugins/store.ts
```diff
@@ -361,11 +366,19 @@ export const store = createStore<FlightCoreStore>({
  * It invokes all Rust methods that are needed to initialize UI.
  */
 async function _initializeApp(state: any) {
-    // Enable dev mode directly if application is in debug mode
-    if (await invoke("is_debug_mode")) {
-        store.commit('toggleDeveloperMode', true);
-    } else {
-        // Disable context menu in release build.
+    // Display dev view if dev mode was previously enabled or if debug mode is active.
+    // Menu style is only modified on debug mode.
```
GeckoEidechse, line 370: Don't forget to update the comment ;)

GeckoEidechse (approved):
Confirmed working in testing. Ready to merge as soon as last comment is resolved ^^

src-vue/src/plugins/store.ts
Alystrasz: Addressed in 7236093389910eda2bb2cbcb9700b35a0ea14b7a

PR #267, opened 2023-04-16
feat: Allow installing plugins

catornot:
not sure about the whole thing.

also I merged #266 because I can't wait


GeckoEidechse:
Is there a plugin on Thunderstore yet to test this against? ^^

catornot:
> Is there a plugin on Thunderstore yet to test this against? ^^

[kyurid](https://northstar.thunderstore.io/package/cat_or_not/kyurid/)

GeckoEidechse:
As @ASpoonPlaysGames said [here](https://github.com/R2NorthstarTools/VTOL/commit/b074ad4edd9405b0acc2f39cb6e886a46e2697de#commitcomment-109575574), a warning whenever a plugin is installed would be nice.

So like when you install a mod with plugin, it should show some pop-up requiring install confirmation (next to limiting plugin install to dev mode for now)

Thing is, I have no idea how to go Vue/TS -> Rust -> Vue/TS -> Rust without getting super ugly, so not sure if I can help you on that part :/

(Still haven't gotten around to testing btw, sorry about that).

src-tauri/src/mod_management/mod.rs
```diff
@@ -370,24 +373,40 @@ pub async fn fc_download_mod_and_install(
     );
 
     // Download the mod
-    let f = match thermite::core::manage::download_file(download_url, path.clone()) {
-        Ok(f) => f,
-        Err(e) => return Err(e.to_string()),
-    };
+    let f = thermite::core::manage::download_file(download_url, path.clone())
+        .map_err(|err| err.to_string())?;
```
GeckoEidechse, line 377: This might sound a bit like bike shedding but I highly suggest to make additional PRs with small changes like these that technically are not part of the main PR.

Cause then I can merge them early and reduce the size of the diff of this PR which in turn makes reviewing this PR easier ^^

src-tauri/src/plugin_management/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+use std::{
+    fs::{self, File, OpenOptions},
+    io,
+    path::PathBuf,
+};
+use tauri::async_runtime::Mutex;
+use thermite::{core::utils::TempDir, prelude::ThermiteError};
+use zip::ZipArchive;
+
+static CAN_INSTALL_PLUGINS: Lazy<Mutex<bool>> = Lazy::new(|| Mutex::new(false));
```
GeckoEidechse, line 12: It might make more sense to pass the `can_install_plugins` as an argument to `fc_download_mod_and_install` via frontend cause that way there's no need to keep track of state in the backend.

In general I try to keep state in frontend so that backend can act completely stateless which means that locking variables etc can be completely omitted ^^

catornot:
this will be the first and last time I work on the frontend

catornot:
added a warning to the installing process with an option to cancel the install

GeckoEidechse (changes requested):
Hmm, so from quick testing trying to install `kyurid` I noticed that:

- [x] There was a second pop-up after some delay when installing plugin which is a bit confusing. Better would be to disable/enable the _OK_ button (UI development is pain)
- [ ] Because the Thunderstore mod only ships a plugin but no Northstar mod, FlightCore is unable to detect it as installed. Not sure how to solve that. Maybe we need to rework something in FlightCore, maybe we should rework plugin spec a bit
- [ ] Enable plugin install, going to another view, and going back to DevView, shows it as disabled again. To resolve, you can add a new variable to `FlightCoreStore` tracking the state (https://github.com/R2NorthstarTools/FlightCore/blob/main/src-vue/src/plugins/store.ts#L61-L87). Might even wanna persist it but don't worry to much about that part, we can look at that together later ^^

catornot:
so now
1. the warning popups don't overlay but I did remove the 5 second wait since I have no idea how to implement properly.
2. the disabled/enabled text in DevView updates correctly now 

for the plugins not showing up as installed. I am not sure what would be the best way but I have some ideas.
1. install a dummy mod in `mods` with data about plugins that came with it
2. add a file to `plugins` folder to keep track of plugins's origin

GeckoEidechse:
CI failure on Windows is likely unrelated to this PR and cause by Rust toolchain update, see also #280

src-vue/src/plugins/store.ts
```diff
@@ -474,6 +479,10 @@ function _initializeListeners(state: any) {
         state.player_count = evt.payload.Ok[0];
         state.server_count = evt.payload.Ok[1];
     });
+
+    listen("display-plugin-warning", async function (evt: TauriEvent<any>) {
+        await display_plugin_warning() // could also disaplay the names of the plugins
```
GeckoEidechse, line 484:
```suggestion
        await display_plugin_warning() // could also display the names of the plugins
```

src-vue/src/plugins/store.ts
```diff
@@ -502,3 +511,19 @@ async function _get_northstar_version_number(state: any) {
             state.northstar_state = NorthstarState.INSTALL;
         })
 }
+
+async function display_plugin_warning() {
+    // just wait for the end user to click ok button
+    await ElMessageBox.alert("This mod contains a plugin. Plugins CAN BE REALLY DANGEROURS since they have to access to your system!", "Warning");
```
GeckoEidechse, line 517: This should be localised. See https://github.com/R2NorthstarTools/FlightCore/blob/main/docs/DEVELOPMENT.md#internationalization on how to do it.

You only need to do the English version. @Alystrasz and I can do French/German respectively ^^

src-vue/src/plugins/store.ts
```diff
@@ -502,3 +511,19 @@ async function _get_northstar_version_number(state: any) {
...
+
+
+    // I switched comfirm and cancel button so end users don't spam throught the popups 
+    // without reading and install a plugin by accident
+    await ElMessageBox.confirm("Do you still want to install this mod with a plugin?", "Warning", { confirmButtonText: "No", cancelButtonText: "Yes", type: "warning", })
```
GeckoEidechse, line 522: Ditto

src-tauri/src/main.rs
```diff
@@ -34,18 +36,23 @@ use mod_management::{
...
 use thunderstore::query_thunderstore_packages_api;
 
-use tauri::{Manager, Runtime};
+use tauri::{AppHandle, Manager, Runtime};
 use tokio::time::sleep;
 
 #[derive(Default)]
 struct Counter(Arc<Mutex<i32>>);
 
+pub static APP_HANDLE: OnceCell<AppHandle> = OnceCell::new();
```
GeckoEidechse, line 54: My Rust knowledge is a bit limited. This essentially sets a global static var, right?

I really could've used this when implementing showing download progress updates in front-end, haha ^^"

GeckoEidechse (changes requested):
Another quick skim.

As for forced timer on popup, I might poke Alystrasz to implement it later... :>

src-tauri/src/main.rs
catornot: yes, it gets initialized here btw https://github.com/R2NorthstarTools/FlightCore/blob/5fb21077e40daca339ae917d7efa9cd1c19a08ec/src-tauri/src/main.rs#L118

GeckoEidechse:
I fixed CI issue on `main` now btw, so just update with `main` to get it resolved ^^

src-vue/src/plugins/store.ts
catornot: this is a comment

catornot:
lol when I tried to fix the merge conflicts it didn't find any.

src-vue/src/plugins/store.ts
```diff
@@ -502,3 +511,29 @@ async function _get_northstar_version_number(state: any) {
...
+            type: "warning",
+        }
+    )
+        .then(() => {
+            invoke("receive_install_status", { comfirmedInstall: false })
+        })
+        .catch(() => {
+            invoke("receive_install_status", { comfirmedInstall: true })
+        })
+}
```
GeckoEidechse, line 539: You stole my newline >:(

catornot: I stole itself, I didn't do anything

GeckoEidechse (changes requested):
Based on #288 the folder plugin is put in should be the Thunderstore mod string of the plugin the mod belongs to, i.e. `R2Northstar/plugins/cat_or_not-kyurid-0.1.0/kyurid.dll`

However testing right now it was placed in `R2Northstar/plugins/kyurid/kyurid.dll`

Also `manifest.json` is also placed in that folder which is probably fine though, depending on what the others say in #288.

When plugin install is disabled it still created folder in plugins folder and placed `manifest.json` which should not happen IMO. If plugin is install is disabled or cancelled it shouldn't still create a folder in `plugins/` folder. ^^"

Also I just realised that we'll have to deal with deleting old plugins on install in case the install is actually an update ^^

catornot:
now the folder is using the thunderstore mod string. 
also the early folder creation was an accident and it is now fixed.

GeckoEidechse (changes requested):
So currently this now installs plugin into `cat_or_not-kyurid` directory

image

Based on #288 I would have assumed it uses full Thunderstore modstring including version number, i.e. `cat_or_not-kyurid-0.1.0` ^^"

Also there's no indication that a mod contained a plugin that was skipped when plugin install is disabled. Maybe a pop-up or warning toast that plugin install was skipped would be nice to avoid "I installed plugin but it's not working" kinda tickets ^^

catornot:
I am not including the version because I would be easier when updating the plugin

Also plugins shouldn't be silently not installing

GeckoEidechse:
> I am not including the version because I think it would be easier when updating the plugin

Eh not really. You technically still have to delete the old mod plugin folder cause the author could've changed the binary name and so simply overriding it might not be enough.

And at that point, whether you delete `cat_or_not-kyurid` or `cat_or_not-kyurid-*` doesn't matter much anymore ^^

> Also plugins shouldn't be silently not installing

Agreed yet that's what it does rn :P

catornot:
> Eh not really. You technically still have to delete the old mod plugin folder cause the author could've changed the binary name and so simply overriding it might not be enough.
>
> And at that point, whether you delete cat_or_not-kyurid or cat_or_not-kyurid-* doesn't matter much anymore ^^

it does. I would to iterate over the whole plugins folder and parse each folder vs just going straight to a folder.

since you want versions to be included so much, I will another iter chain. 

> Agreed yet that's what it does rn :P

```rust
// warn user
if !plugins.is_empty() {
    // check here instead if we can install plugins so people don't get broken mods without plugins
    if !can_install_plugins {
        Err(ThermiteError::MiscError(
            "plugin installing is disabled; this mod contains a plugin; plugins can be enabled in the dev menu".to_string(),
        ))?
    }
// snip //
```

if there is a plugin in the plugins folder and plugin installing is disabled then it will propagate this `MiscError` to the user since this error is getting caught

also there is no mod w/ plugin on thunderstore to test that :P

GeckoEidechse (commented):
Worked in testing, code review tomorrow <3

GeckoEidechse:
In the meantime, don't forget to update after #295 got merged ^^

src-tauri/src/plugin_management/download.rs
```diff
@@ -0,0 +1,162 @@
...
+    game_install: &GameInstall,
+    zip_file: &File,
+    thunderstore_mod_string: &str,
+    can_install_plugins: bool,
+) -> Result<(), ThermiteError> {
+    let plugins_directory = PathBuf::new()
+        .join(&game_install.game_path)
+        .join("R2Northstar")
+        .join("plugins");
+    let temp_dir = TempDir::create(plugins_directory.join("___flightcore-temp-plugin-dir"))?;
```
GeckoEidechse, line 45: Any reason you create a second temp dir instead of using the existing one? ^^

src-tauri/src/plugin_management/download.rs
```diff
@@ -0,0 +1,162 @@
...
+    let manifest_path = temp_dir.join("manifest.json");
+    let mut archive = ZipArchive::new(zip_file)?;
+
+    let parsed_mod_string = ParsedThunderstoreModString::from_str(thunderstore_mod_string).map_err(|_|ThermiteError::MiscError("Gecko why is this returning nothing? anyway the error is failed to parse thunderstore string lmao".into()))?;
```
GeckoEidechse, line 49: I assume this will be changed with #295 merged? :P

src-tauri/src/plugin_management/download.rs
catornot: to unzip the zip

src-tauri/src/plugin_management/download.rs
catornot: yes

src-tauri/src/plugin_management/download.rs
GeckoEidechse: Yeah but like couldn't you do that in the existing temp dir?

GeckoEidechse: [`___flightcore-temp-download-dir/`](https://github.com/R2NorthstarTools/FlightCore/pull/267/files#diff-e8d845933a9237c520ef224431b15bb0d5c9870431d41f06625e2663206371c3R372) should still exist while plugin is being installed ^^

catornot: that's not a temp dir, it never gets cleaned up

GeckoEidechse: > it doesn't when an error is returned

Which should probably get fixed...

>  and also the mod installing code also uses its own temp dir.

Yeah, there's also a bunch of inconsistencies that need to be addressed (and are addressed on `libthermite` version bump IIRC).


Basically, if you can make it work, use existing temp dir, if it doesn't work, make a comment note that that's a temporary solution cause in the long run we only want a single temp dir so that it's easy to clean up by hand on errors etc

GeckoEidechse: Actually, you could just stick your temp dir into the other one as a subdir with little to no modification on your code needed, no?

catornot: ok, this will be temporary

GeckoEidechse (changes requested):
Looks like something broke when merging recent main ._.

image

catornot:
> Looks like something broke when merging recent main ._.

should be fixed now

src-vue/src/i18n/lang/en.json
```diff
@@ -161,5 +162,12 @@
             "Northstar": "Northstar",
             "NorthstarReleaseCandidate": "Northstar release candidate"
         }
+    },
+
+    "plugins": {
+        "warning_dialog": {
+            "warning_text": "This mod contains a plugin. Plugins CAN BE REALLY DANGEROURS since they have to access to your system!",
```
Alystrasz, line 169: Intentional typo? :smile:

src-vue/src/plugins/store.ts
```diff
@@ -48,6 +49,8 @@ export interface FlightCoreStore {
 
     // user custom settings
     mods_per_page: number,
+
```
Alystrasz, line 52:
```suggestion
```

No need for a new line here.

Alystrasz (changes requested):
Plugins don't appear as "installed" on online mods view, while they correctly appear as Thunderstore mods on local view:

image

image

---

Otherwise, features seems to be working correctly, I could install the Furnace plugin on my Ubuntu 23.04.

I however couldn't launch the game:
```text

[SCRIPT UI] [errr] [SCRIPT UI] [error] COMPILE ERROR Undefined variable "FurnaceGetGrid"
[SCRIPT UI] [errr] [SCRIPT UI] [error] ui_furnace.nut line [11] column [48]
```
(you probably should add the mod repository URL to your plugin readme)

src-vue/src/i18n/lang/en.json
catornot: I can't find the typo. :/

src-tauri/src/mod_management/mod.rs
```diff
@@ -432,17 +436,38 @@ pub async fn fc_download_mod_and_install(
...
-    match thermite::core::manage::install_mod(
+    let result_mod = match thermite::core::manage::install_mod(
         author,
         temp_file.file(),
         std::path::Path::new(&mods_directory),
     ) {
-        Ok(()) => (),
-        Err(err) => return Err(err.to_string()),
+        Ok(()) => Ok(()),
+        err if matches!(err, Err(ThermiteError::PrefixError(_))) => err, // probably happens when there is not mod folder found
```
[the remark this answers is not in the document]
catornot: no more probably

src-vue/src/i18n/lang/en.json
GeckoEidechse: > `DANGEROURS`

should be

> `DANGEROUS`

catornot: fixed by [65c8997](https://github.com/R2NorthstarTools/FlightCore/pull/267/commits/65c89977cef06fff654ecb8f9b95369ee467ea95)

catornot:
more conflicts >:( 
and also it would quite not pog if this doesn't get merged before soon:tm: plugins v3 pr

GeckoEidechse:
> more conflicts >:(

Had quite some code refactoring going on the last week or so but should be better now ^^
Just resolve merge conflicts and we can do another round of review

src-tauri/Cargo.lock
```diff
@@ -3951,9 +3979,9 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "steamlocate"
-version = "1.2.1"
```
GeckoEidechse, line 3954: Uh, why is steamlocate being bumped down a version?

GeckoEidechse: Also a bunch of other seemingly unrelated package versions changing, maybe reset `Cargo.lock` to version in `main` and then add dependency again?

src-tauri/src/plugin_management/download.rs
GeckoEidechse: This is still open ig

src-tauri/src/plugin_management/download.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    for file in plugins {
+        fs::copy(file.path(), this_plugin_dir.join(file.file_name()))?;
+    }
+
+    Ok(())
+}
+
+#[tauri::command]
+pub fn receive_install_status(
```
GeckoEidechse, line 156: Doc comment please <3

src-tauri/src/plugin_management/download.rs
catornot: I can wait until plugins v3

catornot:
I don't want to deal with this anymore

PR #270, opened 2023-04-16
fix: Fix incorrect is_empty check

### LINKED ISSUE #269: Removing Thunderstore mod broken
Tested two mods
- `Khalmee-Fall_Damage-1.0.0`
-  `Lawnmower-SpyCloak-1.1.2`

and both could be installed but not uninstalled with FlightCore. Need to investigate the cause for this.

image

GeckoEidechse:
Fixes a regression introduced in #244 that prevented any Thunderstore mods from being uninstalled

Closes #269

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -480,7 +480,7 @@ pub fn delete_thunderstore_mod(
         }
     }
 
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Merging myself cause I confirmed it as working in testing and the whole PR is just inverting the check to enter the branch.

(Typing this up took longer than fixing the bug)

PR #273, opened 2023-04-17
build: Additionally upload binaries separately from CI

GeckoEidechse:
Additionally uploading binaries separately allows for downloading just those files for testing which results in faster download speeds especially on bad internet.

When testing some PR for some feature, debug information generally isn't needed so being able to download the binary without debug information speeds up the process.

Also allows for linking directly to the zip with just the binary as opposed to the one also including debug information which helps when asking non-developers to test builds.


GeckoEidechse (commented):
CI artifacts look as expected with both full zip with debug symbols and zip with just the binaries uploaded.

PR #274, opened 2023-04-17
fix: Update regex to match new release notes format

### LINKED ISSUE #194: Changed release note format in Northstar breaks FlightCore release note page
Going from `#NUMBER` to `REPO#NUMBER` in Northstar release notes broke the ones for FlightCore. Clicking a link in FlightCore on the new format will now open the corresponding page within FlightCore instead of opening the link in the browser like it is supposed to.

image

Alystrasz:
Updates the regex so it matches both old and new release notes formats used to link to Northstar GitHub PRs.

Here's a testing screenshot with regex101 (it shows that new regex matches both formats):
Capture d’écran de 2023-04-17 22-16-39

Closes #194.


GeckoEidechse (approved):
Confirmed working in testing. Both old and new PR links as well as contributor profile links open in external browser as expected.

Many thanks for the PR <3

PR #276, opened 2023-04-19
refactor: Reorder DevView elements

GeckoEidechse:
Moves release note generator to the bottom and mod install section also further down as these tools are used less often than others.

image


Alystrasz (approved):
Successfully tested on Windows, LGTM.

PR #281, opened 2023-04-21
fix: Update game-scanner to avoid deprecated lib

GeckoEidechse:
This is a hotfix until upstream is updated or we replace the library entirely.


GeckoEidechse:
If CI passes this should be good to merge ^^

PR #285, opened 2023-04-22
feat: Show error message if WebView2 not installed

### LINKED ISSUE #181: Display error message if unable to launch WebView2 on Windows
Most comment crash according to sentry is lack of WebView2 on Windows. The FlightCore installer should also install WebView2 if missing but there are edge (hah) cases where the user cancels the install step or purposefully crippled their Edge install.

In such case from the user's perspective the app simply does not launch. If we could at least display an error dialogue, we could tell the user what to do to fix it.

GeckoEidechse:
If user on Windows somehow manages to install FlightCore without WebView2 or decides to uninstall WebView2 again FlightCore won't run.
(Linux is not affected cause WebKitGTK is bundled with the AppImage so running without it would be an impressive feat xD)

In such a case we want to show an error message instructing them how to remedy the issue.

For testing, simply uninstall WebView2 on Windows and then run FlightCore. It should show the following error message

image

TODO:

- [x] Code cleanup
- [x] Fix CI
- [x] Link to right docs page
- [x] Requires #286
- [x] Add screenshot of error message to the troubleshooting docs

Closes #181


ASpoonPlaysGames (approved):
Works in testing, error message shows properly and directs me to the correct troubleshooting section

PR #294, opened 2023-04-28
Plugins detection and disaplying 

catornot:
waiting for #267 

this adds plugins to the local mods area and doesn't support disabling and deleting of plugins lol

image


GeckoEidechse (changes requested):
Did a quick test and it didn't detect the plugin. Is it still checking for only the author and mod part of the folder name? ^^"

catornot:
sounds like a skill issue 
image

GeckoEidechse:
> sounds like a skill issue

Ooooh I was still checking in the ThunderstoreModsView xD

At least it didn't detect it as installed there :P

catornot:
same here

PR #295, opened 2023-04-29
refactor: Use error in `ParsedThunderstoreModString::FromStr` instead of panic

catornot:
that's it

kinda need this for #267


GeckoEidechse (approved):
Looks better than previous, thanks <3

Didn't test but converting `version` from `Option<String>` to `String` should be fine as all places that used the `ParsedThunderstoreModString` type used the parse function and unwrapped it anyway :clueless:

PR #298, opened 2023-04-30
fix: Fetch multiple pages from GitHub PR api

### LINKED ISSUE #297: PR testing tool doesn't fetch all available PRs
Doesn't fit on a screenshot but the oldest mods PR I can install is R2Northstar/NorthstarMods#536. I have no issues for R2Northstar/NorthstarLauncher, maybe because there aren't enough open PRs

GeckoEidechse:
Previously we only fetched the first page which caused us to miss older pull requests.

Closes #297


GeckoEidechse (commented):
@uniboi confirmed working in testing

PR #301, opened 2023-04-30
feat: Launch arguments

### LINKED ISSUE #147: Users cannot specify or edit launch args using FlightCore
Other mod managers can do this, so why not?
(could just edit ns_startup_args, or not, don't think it matters all that much)

Alystrasz:
This interface lists launch arguments (stolen from [wiki page](https://r2northstar.gitbook.io/r2northstar-wiki/using-northstar/launch-arguments)), and enables users to toggle them. A description of each argument can be displayed by hovering related item.

This is merely an interface with the `ns_startup_args.txt` file.

[launch_args_selector.webm](https://user-images.githubusercontent.com/11993538/235377781-c777eb7b-dc38-490a-bc9c-4625b161e631.webm)

Closes #147.

##### TODOs

- [x] `-language` argument
- [x] Should we let users write anything they want as launch args? (I don't think so)
- [x] Can users set up envvars the same way as launch args?
- [x] i18n
- [x] Handle duplicated arguments in file
- [x] Distinction between "official" and "custom" arguments
- [x] Fix clippy issues
- [x] Remove tag animations


Alystrasz:
@GeckoEidechse maybe you'll have answers to some of my questions 😃

GeckoEidechse:
Holy fuck this looks really cool :O

> Should we let users write anything they want as launch args? (I don't think so)

I'm tending towards a yes though in that case it should then also display the raw string in a text box somewhere ^^

src-tauri/src/main.rs
```diff
@@ -145,7 +147,7 @@ fn main() {
             apply_launcher_pr,
             apply_mods_pr,
             get_launcher_download_link,
-            close_application,
+            close_application
```
GeckoEidechse, line 150: You can keep the trailing comma ;)

```suggestion
            close_application,
```

src-vue/src/components/LaunchArgumentsSelector.vue
```diff
@@ -0,0 +1,147 @@
...
+            const officialArguments = [
+                new LaunchArgument("-disablelogs", "Disable logging and creation of log files"),
+                new LaunchArgument("-vanilla", "Disables Northstar loading"),
+                new LaunchArgument("-northstar", "Enables Northstar loading"),
+                new LaunchArgument("-dedicated", "Starts a dedicated server without video output"),
+                new LaunchArgument("-waitfordebugger", "Waits for debugger to connect before launching game"),
+                new LaunchArgument("-enablechathooks", "Enables the use of chathooks for use by mods"),
+                new LaunchArgument("-noplugins", "Disables the plugin system"),
+                new LaunchArgument("-novid", "Disables startup videos"),
+                new LaunchArgument("-nosound", "Disables all game sounds")
```
GeckoEidechse, line 54: Should we localise this?

GeckoEidechse: Okay looking at the naming of the member I guess that's the plan ^^"

src-tauri/src/northstar/launch_arguments.rs
```diff
@@ -0,0 +1,34 @@
...
+    Ok(data.split_whitespace().map(|arg| arg.to_string()).collect())
+}
+
+#[tauri::command]
+pub fn set_launch_arguments(game_path: &str, arguments: Vec<String>) -> Result<(), String> {
+    // If file does not exist, we return an empty list
+    let launch_args_path = format!("{}/ns_startup_args.txt", game_path);
+    if !std::path::Path::new(&launch_args_path).exists() {
+        return Err("Launch arguments file does not exist.".to_string());
+    }
```
GeckoEidechse, line 25: I think we should just create it if it doesn't exist yet ^^

src-vue/src/components/LaunchArgumentsSelector.vue
```diff
@@ -0,0 +1,147 @@
...
+        saveLaunchArgumentsToFile() {
+            const newArgs = this.arguments
+                .filter((value: LaunchArgument, index: number) => {
+                    return this.values[index];
+                })
+                .map((value: LaunchArgument) => value.argumentName)
+
+            invoke<string[]>("set_launch_arguments", {
+                gamePath: this.$store.state.game_path, arguments: newArgs
+            });
```
GeckoEidechse, line 88: Don't forget to handle the promise rejection on error ;)

GeckoEidechse (commented):
Probably a bit too early to do a review atm but I was in the mood ^^"

In regards to clippy errors, half of them are caused by "unused function" which makes no sense to me cause clearly the function is used in the macro to set up tauri. Maybe one of the Rust wizards (cat, emerald, snagg, ...) in the Northstar Discord have an idea. For the rest feel free to poke me if you need/want any help <3

src-tauri/src/main.rs
Alystrasz: I might have removed it by mistake while trying to solve the clippy errors.
I just couldn't import both methods like others...

GeckoEidechse: Yeah, I noticed that while trying as well. Super weird :/

src-tauri/src/northstar/launch_arguments.rs
Alystrasz: Addressed in 25d32c0fe1f05ec3766bf4c433efa56d8b04488d

src-vue/src/components/LaunchArgumentsSelector.vue
Alystrasz: Addressed in a17a3ec0b96f02fe243a0ba38894e770efa356f7

src-tauri/src/main.rs
GeckoEidechse: > I just couldn't import both methods like others...

For some reason importing it this way worked for me

https://github.com/R2NorthstarTools/FlightCore/pull/314/files#diff-2f5e0a90d4195e9986f5e24928dce16b59a80a2cf30f7059b38d55bd7d1eff69R156

GeckoEidechse: I think the `use app::*;` is what's causing the issue. I'll have to look into getting rid of that again

https://github.com/R2NorthstarTools/FlightCore/pull/301/files#diff-2f5e0a90d4195e9986f5e24928dce16b59a80a2cf30f7059b38d55bd7d1eff69R14

src-tauri/src/northstar/launch_arguments.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+#[tauri::command]
+/// This method puts an array of arguments into the ns_startup_args.txt Northstar
+/// launch arguments files.
+/// If the ns_startup_args.txt file does not exist, this will create it.
+pub fn set_launch_arguments(game_path: &str, arguments: Vec<String>) -> Result<(), String> {
+    let launch_args_path = format!("{}/ns_startup_args.txt", game_path);
+    
+    match std::fs::write(launch_args_path.clone(), arguments.join(" ")) {
```
catornot, line 33: consider using `if let` or `map_err` since you are only using the error

catornot (changes requested):

src-tauri/src/northstar/launch_arguments.rs
Alystrasz: Hopefully addressed with 15bdcde17f0f40313d3bc709ae1c535e5fcbeaca

GeckoEidechse (changes requested):
So turns out that Convar arguments are formatted

```
+my_convar 123
```
not
```
+my_convar=123
```

This makes parsing for convar args quiet a bit more tedious.

The easiest solution I can come up with is to just check for the leading `+` and then considering the next argument as part of the current one.

Of course, that still requires writing some parser ._.

catornot:
> Of course, that still requires writing some parser ._.

make a parser using a iter chain :)

> The easiest solution I can come up with is to just check for the leading + and then considering the next argument as part of the current one.

itertools has usefull stuff for this -> [chunks](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.chunks)

Alystrasz:
While FlightCore UI can try preventing users from breaking their launch arguments configuration by adding random words in it, I feel like the rust backend should be responsible for its integrity.
What do you think should happen when backend detects a malformed arguments file? Overwrite it with correctly-formatted arguments? Refuse to parse it and raise an error?

GeckoEidechse:
> What do you think should happen when backend detects a malformed arguments file? Overwrite it with correctly-formatted arguments? Refuse to parse it and raise an error?

Correcting user mistake sounds very difficult to do tbh. It makes more sense IMO to just reject with error and if possible say what's wrong (though the second part of that might also be quite difficult to do ^^)

ASpoonPlaysGames:
Honestly, there shouldn't rly be any handling of this sorta thing. It's almost impossible to know which launch args exist and are valid and stuff. Launch args are for advanced users and they should not need their hand to be held

ASpoonPlaysGames:
To elaborate:
- Any convar (which mods can add) is a valid launch arg (with the + prefix i believe)
- Plugins can just check command line args for whatever they want, and there isn't a manifest or anything for it
- Launcher can do the same as plugins (but we can control that a bit more)

Alystrasz:
Thanks for your feedback!
What will be the correct way to use launch arguments once associated file has been removed (https://github.com/R2Northstar/NorthstarLauncher/pull/517)? Directly as executable arguments?

Alystrasz:
Since we're talking removing the `ns_launch_args.txt` file again, I'll rework this by storing arguments in flightcore itself.

Alystrasz:
image

I need help on the two following points:
- making sure launch args are correctly sent to Steam
- finding out why launch args is interpreted as a string by launcher

src-tauri/src/northstar/mod.rs
```diff
@@ -212,9 +214,10 @@ pub fn launch_northstar(
     {
         let ns_exe_path = format!("{}/NorthstarLauncher.exe", game_install.game_path);
         let ns_profile_arg = format!("-profile={}", game_install.profile);
+        let args_str = &launch_args.join(" ");
 
         let _output = std::process::Command::new("C:\\Windows\\System32\\cmd.exe")
-            .args(["/C", "start", "", &ns_exe_path, &ns_profile_arg])
+            .args(["/C", "start", "", &ns_exe_path, &ns_profile_arg, args_str])
```
ASpoonPlaysGames, line 220: Isn't this what is causing the args to be interpreted as one string? you are adding them as a single argument instead of adding each argument individually

Alystrasz: That was exactly that :')
Addressed in 2a4fc2fbf34f5be3cdeb5031efbb17420a74cfbc

Alystrasz:
This diverged too much from `main`, closing in favour of #1127.

PR #304, opened 2023-05-05
refactor: Always remove zip files

catornot:
related to https://github.com/R2NorthstarTools/FlightCore/pull/267#discussion_r1186120176

to delete the folder instead of files, the function would need a rewrite since the implementation rn doesn't allow it

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -67,9 +67,7 @@ impl TempFile {
 
 impl Drop for TempFile {
     fn drop(&mut self) {
-        println!("path {}", self.1.display());
-
-        _ = dbg!(fs::remove_file(&self.1));
+        _ = fs::remove_file(&self.1)
     }
 }
 
@@ -361,7 +359,7 @@ pub async fn fc_download_mod_and_install(
     };
 
     // Recursively install dependencies
-    for dep in dbg!(deps) {
+    for dep in deps {
         match fc_download_mod_and_install(game_install, &dep).await {
             Ok(()) => (),
             Err(err) => {
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -361,6 +361,7 @@ pub async fn fc_download_mod_and_install(
         Ok(deps) => deps,
         Err(err) => return Err(err.to_string()),
     };
+    dbg!(deps.clone());
 
     // Recursively install dependencies
     for dep in deps {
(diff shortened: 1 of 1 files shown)


src-tauri/src/mod_management/mod.rs
```diff
@@ -331,10 +359,9 @@ pub async fn fc_download_mod_and_install(
         Ok(deps) => deps,
         Err(err) => return Err(err.to_string()),
     };
-    dbg!(deps.clone());
 
     // Recursively install dependencies
-    for dep in deps {
+    for dep in dbg!(deps) {
```
GeckoEidechse, line 364: Leftover debug print

src-tauri/src/mod_management/mod.rs
```diff
@@ -53,6 +52,35 @@ pub struct ModJson {
...
+    pub fn file(&self) -> &fs::File {
+        &self.0
+    }
+}
+
+impl Drop for TempFile {
+    fn drop(&mut self) {
+        println!("path {}", self.1.display());
+
+        _ = dbg!(fs::remove_file(&self.1));
```
GeckoEidechse, line 72: Maybe use [`log`](https://crates.io/crates/log) crate related functions (`log::info!()` etc) for prints so that we can keep stuff consistent ^^

src-tauri/src/mod_management/mod.rs
```diff
@@ -53,6 +52,35 @@ pub struct ModJson {
     version: Option<String>,
 }
 
+#[derive(Debug)]
+pub struct TempFile(fs::File, PathBuf);
```
GeckoEidechse, line 56: Just so that I understand the code correctly. The point of this struct is that when the variable is dropped, the destructor is called which deletes the file again, right?

If that's that case maybe add a comment mentioning the purpose/use-case of this struct ^^

catornot: isn't it obvious -.-

src-tauri/src/mod_management/mod.rs
catornot: I forgot to remove debug stuff

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Took me a sec, I'm a Rust noob after all, so it would be appreciated <3

catornot: not needed imo

GeckoEidechse: Push perms go brrrrrr 0f636df394051f777bf06ed16978963a1212862c

GeckoEidechse (approved):
Confirmed working in testing by installing some mod and observing temp file getting created and then again deleted.

PR #305, opened 2023-05-05
refactor: Prepare code for installing older Northstar versions

GeckoEidechse:
This PR aims to refactor Northstar installation code to make it easier to later add functionality to install a specific Northstar versions

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -280,6 +280,7 @@ pub async fn install_latest_northstar(
     window: tauri::Window,
     game_path: &str,
     northstar_package_name: String,
+    version_number: Option<String>,
 ) -> Result<String, String> {
     let index = thermite::api::get_package_index().unwrap().to_vec();
     let nmod = index
@@ -288,11 +289,14 @@ pub async fn install_latest_northstar(
         .ok_or_else(|| panic!("Couldn't find Northstar on thunderstore???"))
         .unwrap();
 
+    // Use passed version or latest if no version was passed
+    let version = version_number.unwrap_or_else(|| nmod.latest.clone());
+
     log::info!("Install path \"{}\"", game_path);
 
     match do_install(
         window,
-        nmod.versions.get(&nmod.latest).unwrap(),
+        nmod.versions.get(&version).unwrap(),
         std::path::Path::new(game_path),
     )
     .await
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -281,7 +281,6 @@ pub async fn install_latest_northstar(
     game_path: &str,
     northstar_package_name: String,
 ) -> Result<String, String> {
-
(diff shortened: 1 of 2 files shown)


GeckoEidechse:
@catornot you wanna review this PR? At least all the changes in here are backend/Rust stuff and AFAIK you have the most Rust knowledge + FlightCore knowledge combo out of the people I know ^^

src-tauri/src/lib.rs
```diff
@@ -279,31 +279,24 @@ async fn do_install(
...
-
     let index = thermite::api::get_package_index().unwrap().to_vec();
     let nmod = index
         .iter()
         .find(|f| f.name.to_lowercase() == northstar_package_name.to_lowercase())
         .ok_or_else(|| panic!("Couldn't find Northstar on thunderstore???"))
         .unwrap();
 
+    // Use passed version or latest if no version was passed
+    let version = version_number.unwrap_or_else(|| nmod.latest.clone());
```
catornot, line 293: this could be a borrow instead to avoid cloning and you borrow this anyway later

src-tauri/src/main.rs
```diff
@@ -327,9 +327,23 @@ async fn install_northstar_caller(
     window: tauri::Window,
     game_path: String,
     northstar_package_name: Option<String>,
+    version_number: Option<String>,
 ) -> Result<bool, String> {
     log::info!("Running");
-    match install_northstar(window, &game_path, northstar_package_name).await {
+
+    // Get Northstar package name (`Northstar` vs `NorthstarReleaseCandidate`)
+    let northstar_package_name = match northstar_package_name {
```
catornot, line 335: the match case can be replaced with `map` and `unwrap_or` ;)

catornot (changes requested):
ok

src-tauri/src/lib.rs
GeckoEidechse: So like this a769530b59c0a46813f1ac881dda628267438329 ? ^^"

src-tauri/src/main.rs
GeckoEidechse: So like

```rust
    let northstar_package_name = northstar_package_name
        .map(|name| {
            if name.len() <= 1 {
                "Northstar".to_string()
            } else {
                name
            }
        })
        .unwrap_or("Northstar".to_string());
```
? ^^"

catornot: yeah

src-tauri/src/lib.rs
catornot: yes

src-tauri/src/main.rs
GeckoEidechse: Done in 05483e672b95c664abb1a4df1918006c3b956ebb

catornot (approved):
good

PR #306, opened 2023-05-05
temp: install older northstar version

GeckoEidechse:
This is wip/temp so that I can see the diff more easily. Contents will be spliced out and merged in separate PRs.


GeckoEidechse:
Superseeded by #325

PR #312, opened 2023-05-06
refactor: Put doc comment before macro

GeckoEidechse:
Doc comments should be placed before the procedural macro decorating the function being documented.


Alystrasz (approved):
Thanks for the refactoring.

PR #313, opened 2023-05-06
refactor: Only pass commit sha for installing launcher CI build

GeckoEidechse:
This refactor allows for easier usage later when installing a launcher build based on commit sha only, e.g. for installing newest build of a branch.

Previously one had to pass an entire pull request struct which was unnecessary as only the commit sha from that struct was actually being used.


Alystrasz (changes requested):
You should update documentation, then this will be ready for merge.

Alystrasz (approved):

PR #314, opened 2023-05-06
feat: WIP code to install launcher from git main

GeckoEidechse:
Works but needs some more cleanup still

Requires:
- [ ] #313 

TODO:
- [ ]  Cleanup
  - [ ] Reduce duplicated code
  - [ ] Move code from `main.rs` to right source file

### DIFF
--- a/src-tauri/src/github/mod.rs
+++ b/src-tauri/src/github/mod.rs
@@ -21,8 +21,8 @@ pub struct TagWrapper {
 }
 
 #[derive(Debug, Deserialize)]
-struct CommitInfo {
-    sha: String,
+pub struct CommitInfo {
+    pub sha: String,
     commit: Commit,
 }
 
--- a/src-tauri/src/github/pull_requests.rs
+++ b/src-tauri/src/github/pull_requests.rs
@@ -37,24 +37,24 @@ pub struct PullsApiResponseElement {
 
 // GitHub API response JSON elements as structs
 #[derive(Debug, Deserialize, Clone)]
-struct WorkflowRun {
-    id: u64,
-    head_sha: String,
+pub struct WorkflowRun {
+    pub id: u64,
+    pub head_sha: String,
 }
 #[derive(Debug, Deserialize, Clone)]
-struct ActionsRunsResponse {
-    workflow_runs: Vec<WorkflowRun>,
-}
-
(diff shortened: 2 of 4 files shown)


src-tauri/src/development/mod.rs
```diff
@@ -0,0 +1,81 @@
+use crate::github::{
+    pull_requests::{check_github_api, download_zip_into_memory, get_launcher_download_link},
+    CommitInfo,
+};
+
+#[tauri::command]
+pub async fn install_git_main(game_install_path: &str) -> Result<String, String> {
+    let url = "https://api.github.com/repos/R2Northstar/NorthstarLauncher/commits";
```
Alystrasz, line 8: Shouldn't this be in the constants package?

src-tauri/src/development/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+        Err(err) => {
+            return Err(format!("Failed unzip: {}", err));
+        }
+    };
+
+    // Copy only necessary files from temp dir
+    // Copy:
+    // - NorthstarLauncher.exe
+    // - Northstar.dll
+    let files_to_copy = vec!["NorthstarLauncher.exe", "Northstar.dll"];
```
Alystrasz, line 50: Are we sure this list will never change?

GeckoEidechse: Can be sure? No.
Has it ever changed in Northstar's development? Also no :P

src-tauri/src/development/mod.rs
GeckoEidechse: Done with 023fee1ea15a6b44cea4fe0aa878a45f95b0d385

src-tauri/src/development/mod.rs
Alystrasz: Fine by me :)

Alystrasz:
Could you update Northstar version after install?

GeckoEidechse:
> Could you update Northstar version after install?

Doesn't work cause FlightCore only reads it from the `Northstar.XXXX` mods' `mod.json`s which not changed when changing just the launcher binary.

I want to also read version number from the `.dll` and `.exe` but I have no idea how to (and also that should be a separate PR) ^^

Alystrasz (approved):

PR #315, opened 2023-05-06
docs: Mention weblate for translating

GeckoEidechse:
Adds mention of weblate to README to tell users how they can submit translations.

Keeping as draft until weblate is fully set up.


GeckoEidechse:
@p0358 so everything weblate related works on my end. Anything that we still need to setup on the Harmony weblate before I merge docs for it? ^^

p0358:
I think it should be ready, can't think of anything else to setup

Alystrasz (approved):
LGTM.
Thanks for the documentation (and thanks @p0358 for the weblate setup).

PR #319, opened 2023-05-06
chore: Bump steamlocate to 1.2.1

### LINKED ISSUE #318: `steamlocate` `v1.2.0` panicking causing thread crash
From sentry:

image

`panic` is caused when parsing fails. Shouldn't happen if user has Steam version of Titanfall2 installed.

Should be fixed by updating to `v1.2.1`

GeckoEidechse:
Closes #318


GeckoEidechse:
Note sure if there's even anything to review here. I'll just merge when CI passes.

PR #321, opened 2023-05-07
feat: Enable Polish translations

GeckoEidechse:
Adds _Polish_ into the language dropdown menu in settings.

image


GeckoEidechse:
@p0358 just to be clear, _"Polish"_ in Polish is _"Polski"_, right? ^^

p0358:
Yes, but it should be not capitalized, aka "polski". We always write adjectives non-capitalized. It might look a bit like a sore thumb, but going down the line it's not gonna be the only language with such rule (for example Russian is the same in this regard iirc)

Alystrasz (approved):

PR #322, opened 2023-05-07
fix: Handle no internet on serverbrowser fetch

### LINKED ISSUE #263: Player count load panics when there's not internet
`get_server_player_count()` has unwraps that cause a thread panic when internet connection is gone.

This is not visible to the end-user cause a thread panic does not crash the application but still isn't a good look either.

To reproduce:
1. Launch FlightCore (make sure to do it via console so you actually see the output)
2. remove internet connection
3. thread panic

Part of #33

GeckoEidechse:
Instead of panic, simply return an error.

When there's no connection to master server, either due to it being down or the user's internet connection being offline, there's no need to panic the thread. Rather simply return an error (which is silently logged in frontend)

Closes #263


Alystrasz:
Servers list fetching is not the only place where things go wrong with no Internet connection; could we have like a global error handler that returns an error for all methods requiring an Internet connection?

GeckoEidechse:
> Servers list fetching is not the only place where things go wrong with no Internet connection; could we have like a global error handler that returns an error for all methods requiring an Internet connection?

Hmm not sure how to implement that tbh. The way I'd solve it, is by replacing all `unwrap`s with `?`s and bubble up the error to then handle properly on frontend (maybe even using custom error types as described in #323)

There might also be a way to send a message to frontend on `panic` but that's likely limited in what kind of information it can pass back which would make it difficult to differentiate the cause of error.

Overall there isn't really a _catch all exceptions_ kinda thing as `panic`s are more like a `exit(-1)` kinda deal (though in most cases in FlightCore it's just that thread that panics)


For some reason `/client/servers` reading failing due to lack of connection is one of the primary cause of thread crashes according to sentry. My assumption is that some firewalls are blocking it due to being an "unknown domain name". Hence I'm prioritising this function in the quest of removing all `unwraps`. ^^

PR #324, opened 2023-05-08
feat: Add category in changelog for i18n

GeckoEidechse:
When auto generating changelog from commit titles, this turns `i18n` (which is what we set weblate PRs to be prefixed with) into `Translations`

So

```
i18n: Add Polish translations via Weblate (#320)
```
becomes
```
**Translations:**
- Add Polish translations via Weblate (#320)
```

For testing, one can simply generate changelog between `v1.14.1` and `v1.15.0` and look at the generated changelog.

Note that I'm well aware _i18n_ stands for _Internationalization_ but _Translations_ sounds more sensible to me in this case ^^

### DIFF
--- a/src-tauri/src/constants.rs
+++ b/src-tauri/src/constants.rs
@@ -33,8 +33,8 @@ pub const TITANFALL2_ORIGIN_IDS: [&str; 2] = ["Origin.OFR.50.0001452", "Origin.O
 pub const TITANFALL2_STEAM_ID: &str = "1237970";
 
 // Order in which the sections for release notes should be displayed
-pub const SECTION_ORDER: [&str; 9] = [
-    "feat", "fix", "docs", "style", "refactor", "build", "test", "chore", "other",
+pub const SECTION_ORDER: [&str; 10] = [
+    "feat", "fix", "docs", "style", "refactor", "build", "test", "i18n", "chore", "other",
 ];
 
 // GitHub API endpoints for launcher/mods PRs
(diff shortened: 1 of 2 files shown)


Alystrasz:
Do you think all i18n-related changes should go under this new category? (I do)

image

GeckoEidechse:
> Do you think all i18n-related changes should go under this new category? (I do)

Sorta? Depends on the change ig.

Are you saying I should change the category name or keep it? ^^

Alystrasz:
I'm saying all i18n-related changes (*"Enable Polish translation #321"* for instance) should be featured under this section :)

GeckoEidechse:
> I'm saying all i18n-related changes (_"Enable Polish translation #321"_ for instance) should be featured under this section :)

Then they just need to be committed with `i18n:` prefix instead of `feat:` as I [only check the commit prefix](https://github.com/R2NorthstarTools/FlightCore/pull/324/files#diff-ce9cf27881ddb8aa3ac5bb11c923daaaeed5521d960a371c94343689f1fc8619R160-R168). Don't think it makes sense to add a whole natural language parser or similar ^^

Also not gonna rewrite commit message for past commits :P

Alystrasz:
> Then they just need to be committed with i18n: prefix

We gotta be careful regarding this then :)

GeckoEidechse:
> > Then they just need to be committed with i18n: prefix
> 
> We gotta be careful regarding this then :)

I can always manually edit changelog before releasing if need be :P
Also weblate is already setup to PR with `i18n:` prefix :D

PR #325, opened 2023-05-08
feat: Add button to install older Northstar version in DevView

GeckoEidechse:
Shows all packages on Thunderstore and allows selecting/installing older ones

image

TODO:

- [x] Screenshot showcase :eyes: 
- [x] ~~Move stuff out of `main.rs`~~ delayed for #329 
- [x] General code cleanup


Alystrasz (approved):
Code is clean, feature tested successfully on Ubuntu 23.04.

Alystrasz:
What's the plan with this feature?
Do you want to keep it a dev feature, or eventually move it to play view some day?

GeckoEidechse:
> What's the plan with this feature? Do you want to keep it a dev feature, or eventually move it to play view some day?

Plan would be to keep it a dev-only feature ^^

In general, there shouldn't be a reason for a player to downgrade. Downgrade only makes sense for them in two cases
1. bug in Northstar (but then we should just fix it in Northstar)
2. mod incompatibility (but then mod author should update their mod)

PR #327, opened 2023-05-08
i18n: Add empty translation file for Italian

GeckoEidechse:
This is mostly just for testing to see whether weblate will pick it up automatically


GeckoEidechse:
@p0358 merging this should result in weblate adding _Italian_ as language option, right?

PR #330, opened 2023-05-09
refactor: Move code from `lib.rs` into `main.rs`

### LINKED ISSUE #329: Get rid of `lib.rs`
`lib.rs` is generally used for libraries, as opposed to `main.rs` for binary applications. Given that FlightCore is not intended to be used as library, we should get rid of `lib.rs` entirely as I have reason to believe it's the cause for various import issues when trying to import code from sister modules.

GeckoEidechse:
Doing the remaining refactor in one big chunk as it's simply easier.

There isn't much to code review here as it's just a copy-paste from `lib.rs` into `main.rs`. Main concern is just testing functionality/stability ^^

Closes #329


Alystrasz (approved):
Confirmed working on Ubuntu 23.04; tested features include:
* downloading lastest Northstar version
* downloading ancient Northstar version
* installing mods
* updating language

PR #335, opened 2023-05-09
refactor: Move `check_is_valid_game_path` to `main.rs`

GeckoEidechse:
Moves  `check_is_valid_game_path` into `main.rs`. Ultimately it should land in a submodule but moving it there directly while `lib.rs` still exists causes some weird import/export issues that I don't fully understand yet.

Part of #329


Alystrasz (approved):

PR #337, opened 2023-05-09
Move `linux_checks_librs` to own module

GeckoEidechse:
Part of #329

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -9,9 +9,6 @@ mod platform_specific;
 #[cfg(target_os = "windows")]
 use platform_specific::windows;
 
-#[cfg(target_os = "linux")]
-use platform_specific::linux;
-
 use serde::{Deserialize, Serialize};
 use sysinfo::SystemExt;
 use tokio::time::sleep;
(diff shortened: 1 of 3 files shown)


catornot:
maybe `linux_checks` gets compiled on windows too?

GeckoEidechse:
> maybe `linux_checks` gets compiled on windows too?

Yeah but like both Linux and Windows say unused function... T_T

src-tauri/src/main.rs
```diff
@@ -42,6 +42,8 @@ use mod_management::{
 mod northstar;
 use northstar::get_northstar_version_number;
 
+mod platform_specific;
```
catornot, line 45: you had this before

```
#[cfg(target_os = "windows")]
use platform_specific::windows;


#[cfg(target_os = "linux")]
use platform_specific::linux;
```

src-tauri/src/lib.rs
```diff
@@ -9,9 +9,6 @@ mod platform_specific;
 #[cfg(target_os = "windows")]
 use platform_specific::windows;
```
catornot, line 10: also why wasn't this moved/removed?

src-tauri/src/main.rs
GeckoEidechse: Nope that just gives an additional error of
```
warning: unused import: `platform_specific::linux`
  --> src/main.rs:49:5
   |
49 | use platform_specific::linux;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```
which makes like no sense to me T_T

PR #342, opened 2023-05-10
feat: Make search fuzzy

GeckoEidechse:


GeckoEidechse:
@Alystrasz how do I reuse TypeScript functions between multiple views? ^^

src-vue/src/views/mods/ThunderstoreModsView.vue
```diff
@@ -79,9 +80,11 @@ export default defineComponent({
             return this.mods.filter((mod: ThunderstoreMod) => {
                 // Filter with search words (only if search field isn't empty)
                 const inputMatches: boolean = this.searchValue.length === 0
-                    || (mod.name.toLowerCase().includes(this.searchValue)
-                        || mod.owner.toLowerCase().includes(this.searchValue)
-                        || mod.versions[0].description.toLowerCase().includes(this.searchValue));
+                    || (
+                        fuzzy_filter(mod.name, this.searchValue) ||
+                        fuzzy_filter(mod.owner, this.searchValue) ||
+                        mod.versions[0].description.toLowerCase().includes(this.searchValue)
```
Alystrasz, line 86: No fuzzy filtering on mods' descriptions? :(

src-vue/src/utils/filter.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Implements a fuzzy filter
```
Alystrasz, line 2: While one can understand fuzzy filtering by looking at this implementation, I would like to have some explanations here (maybe including an external documentation link like Wikipedia).

Alystrasz (changes requested):

src-vue/src/views/mods/ThunderstoreModsView.vue
GeckoEidechse: It's cause that would bring up a lot of noise/false positives.

For comparison, results with same search term as PR description but with fuzzy filter on Thunderstore description enabled:
image

src-vue/src/utils/filter.ts
GeckoEidechse: Done in 5ee7a99403f7b8ed6ef9a45ae6e5d1f635dde509

GeckoEidechse:
CI failing due to outdated dependency, ugh

src-vue/src/views/mods/ThunderstoreModsView.vue
Alystrasz: Fair!

Alystrasz (approved):
Comments were addressed, code looks good and feature works as expected on Ubuntu 23.04.

PR #344, opened 2023-05-11
fix: Pass correct argument to `get_launcher_download_link`

### LINKED ISSUE #343: Downloading launcher PRs is broken
image

GeckoEidechse:
Function signature changed in #313 but frontend was never updated.

To test, simply try using the _Download_ button on a Northstar launcher pull request without/with this PR.

Closes #343

### DIFF
--- a/src-vue/src/plugins/modules/pull_requests.ts
+++ b/src-vue/src/plugins/modules/pull_requests.ts
@@ -37,7 +37,7 @@ export const pullRequestModule = {
                 });
         },
         async downloadLauncherPR(state: PullRequestStoreState, pull_request: PullsApiResponseElement) {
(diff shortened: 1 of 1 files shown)


Alystrasz:
What exactly should be done to reproduce the issue?

I can install launcher and mods PRs without a problem using current `main`.

image

GeckoEidechse:
> What exactly should be done to reproduce the issue?
> 
> I can install launcher and mods PRs without a problem using current `main`.

The `Download`, not the `Install` button. I made sure to test `Install` when I did #313 but I forgot about `Download`. Also it only affects launcher ^^

Alystrasz (approved):
I confirm this PR effectively fixes mentioned issue.

PR #345, opened 2023-05-11
refactor: Move NS install code into own module

GeckoEidechse:
Moves the code in question out from `main` into its own module

For testing, simply install/update Northstar (switch between release/release-candidate if no update is pending otherwise).


Alystrasz (approved):

PR #346, opened 2023-05-11
fix: Do not crash if not able to connect to TS

GeckoEidechse:
Removes a bunch of `unwrap`s that would cause the thread to crash if connection to Thunderstore failed.

Fixes https://northstar-kv.sentry.io/issues/4173358275/

Part of #33


Alystrasz (approved):
Confirmed working on Ubuntu 23.04 (warning log correctly appears).
Don't you want to display some error message in UI though? For now, if TS mods fetching failed, users will never know.

GeckoEidechse:
> Don't you want to display some error message in UI though? For now, if TS mods fetching failed, users will never know.

Good catch. Thought we were already doing that. Fixed in fbea5acac69e4ee1174bd02a548d792bf8355cee

PR #347, opened 2023-05-11
refactor: Move some functions to utility module

GeckoEidechse:
Moves `force_panic` and `is_debug_mode` to `utility.rs` to cleanup `main.rs`.

For testing:
- `force_panic`: simply do a force panic via DevView. FlightCore should crash as expected
- `is_debug_mode`: Debug build should show stripes on top bar while non-debug should not show them


Alystrasz (approved):

Issue #351, opened 2023-05-12
Cleanup `main.rs`


GeckoEidechse:
After #330 `main.rs` is a bit too crowded and code should now be moved into dedicated submodules.

GeckoEidechse:
Closed via

- #616 
- #618 
- #619 
- #622
- #623 
- #626 
- #627 
- #628 
- #629

PR #352, opened 2023-05-12
refactor: Move server browser fetch to util module

GeckoEidechse:
To clean up `main.rs`

For testing simply launch FlightCore and see if it can fetch playercount.

Part of #351


GeckoEidechse (commented):
Confirmed working in testing on Windows 10.

PR #357, opened 2023-05-15
chore: Update Vue npm package

GeckoEidechse:
Previous version was pulling in a dependency that is now deprecated, causing CI to fail.


GeckoEidechse (commented):
Seems to still work on Linux.

PR #358, opened 2023-05-15
build: Add dependabot config

GeckoEidechse:
Adds config for running dependabot. AFAIK adding that config should be enough. No other changes needed.

https://github.blog/2022-04-06-how-dependabot-empowers-you-to-keep-your-projects-secure/

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file


GeckoEidechse:
Before merging this I would bump all dependencies to avoid getting a million PRs from dependabot that bumps dependencies.

Question is, @Jan200101 @Alystrasz do you have any open PRs right now that adds/removes dependencies, be it frontend or backend? Cause this PR here can wait tbh and I don't want to introduce any additional conflicts into your PRs ^^

Jan200101:
Got none

GeckoEidechse:
Awesome. #301 also doesn't seem to modify any dependencies so Imma try bumping them again and then merging this PR here ^^

PR #360, opened 2023-05-15
fix: Add missing fuzzy filter for local mods search

GeckoEidechse:
Forgot about that one in #342, whoops (/.-)


GeckoEidechse (commented):
Tested myself on Windows and worked ^^

PR #369, opened 2023-05-25
Add antivirus mention to `Failed to connect to Thunderstore.` error

CooldudePUGS:
This (probably) works 

Can double check later at home today if it's functionality is questionable, but as far as my code knowledge goes `\n` should also work here


GeckoEidechse:
I'm gonna close this PR for now until I have a proper solution ^^

PR #375, opened 2023-06-02
fix: Remove some unwraps for fetching GitHub API

GeckoEidechse:
Remove some unwraps for fetching GitHub API to prevent thread from panicking on no internet

Addresses: https://northstar-kv.sentry.io/share/issue/3ebc5a2c95234439863ad80a5001f203/ (finally found out how to publicly share crash reports)

Part of #33


GeckoEidechse:
Still needs to be handled in frontend https://github.com/R2NorthstarTools/FlightCore/blob/main/src-vue/src/plugins/store.ts#L262

PR #383, opened 2023-06-07
feat: Add basic Proton management

Jan200101:
based on #180 
adds logic for installing, uninstalling and getting the local NSProton version.


Jan200101:
I don't know how to fix the clippy issues because they are completely different locally.

src-tauri/src/main.rs
```diff
@@ -628,3 +631,45 @@ fn launch_northstar_steam(
...
+    match linux::get_local_ns_proton_version() {
+        Ok(ver) => Ok(ver),
+        Err(err) => Err(err),
+    }
+
+    #[cfg(target_os = "windows")]
+    {
+        Err("Not supported on Windows".to_string())
+    }
+}
```
catornot, line 675: same things here

src-tauri/src/platform_specific/linux.rs
```diff
@@ -1,8 +1,78 @@
...
+    // Download NorthstarProton release
+    let archive = thermite::core::download_ns_proton(latest, path.clone())?;
+    dbg!("Download done");
+
+    let compat_dir = get_proton_dir().unwrap();
+    std::fs::create_dir_all(compat_dir.clone())?;
+
+    // Extract to Proton dir
+    thermite::core::install_ns_proton(&archive, compat_dir)?;
+    dbg!("Install done");
```
catornot, line 32: flight core has a logging crate 
```suggestion
    log::info!("Install done");
```
iirc

src-tauri/src/main.rs
```diff
@@ -628,3 +631,45 @@ fn launch_northstar_steam(
...
+
+#[tauri::command]
+fn uninstall_northstar_proton_wrapper() -> Result<(), String> {
+    #[cfg(target_os = "linux")]
+    match linux::uninstall_ns_proton() {
+        Ok(()) => Ok(()),
```
catornot, line 653: this match case be removed (clippy suggested this too)

src-tauri/src/main.rs
```diff
@@ -628,3 +631,45 @@ fn launch_northstar_steam(
...
+        Err(err) => Err(err),
+    }
+
+    #[cfg(target_os = "windows")]
+    {
+        Err("Not supported on Windows".to_string())
+    }
+}
```
catornot, line 661: can probably be replaced by 
```rust
#[cfg(target_os = "windows")]
Err("Not supported on Windows".to_string())
```

saves 2 lines

catornot:
> I don't know how to fix the clippy issues because they are completely different locally.

maybe try updating clippy?

GeckoEidechse:
Thanks a lot for the PR <3

Will hopefully get to testing tomorrow, as for code review stuff, I think @catornot already mentioned most things ^^

> > I don't know how to fix the clippy issues because they are completely different locally.
> 
> maybe try updating clippy?

This. CI pulls in newest Rust version which is currently `1.70`. You can check with e.g. `cargo --version` and then just run `rustup update` to get to newest version ^^

Jan200101:
> maybe try updating clippy?

my local clippy is clippy 0.1.70, this turned out to be some tauri thing that needed poking to make happy.

src-tauri/Cargo.toml
```diff
@@ -67,6 +67,8 @@ open = "3.2.0"
 # Windows API stuff
 winapi = "0.3.9"
 semver = "1.0"
+# easier fs handling
```
GeckoEidechse, line 70:
```suggestion
# easier filesystem handling
```

src-tauri/src/platform_specific/linux.rs
```diff
@@ -1,8 +1,78 @@
...
+    let steam_dir = steamlocate::SteamDir::locate()?;
+    let compat_dir = format!("{}/compatibilitytools.d/", steam_dir.path.display());
+
+    Some(compat_dir)
+}
+
+/// Downloads and installs NS proton
+/// Assumes Steam install
+pub fn install_ns_proton() -> Result<(), thermite::prelude::ThermiteError> {
+    // Get latest NorthstarProton release
```
GeckoEidechse, line 17: Being in a location with terrible internet rn I would prefer if there were `log::info!()`s  before fetching newest NS proton release version and downloading  corresponding zip. At least when testing rn, I was just looking at an essentially blank terminal waiting for something to happen ^^

src-tauri/src/platform_specific/linux.rs
```diff
@@ -1,8 +1,78 @@
...
+    let latest = thermite::core::latest_release()?;
+
+    let temp_dir = env::temp_dir();
```
GeckoEidechse, line 20: I really wish I knew about `env::temp_dir` before implementing my own version of temp dir x_x

Time for me to refactor some code again after this xD

src-tauri/src/platform_specific/linux.rs
```diff
@@ -1,8 +1,78 @@
...
+    let path = format!("{}/nsproton-{}.zip", temp_dir.display(), latest);
+
+    // Download NorthstarProton release
+    let archive = thermite::core::download_ns_proton(latest, path.clone())?;
+    dbg!("Download done");
```
GeckoEidechse, line 25:
```suggestion
    log::info!("NorthstarProton download done");
```

src-tauri/src/platform_specific/linux.rs
```diff
@@ -1,8 +1,78 @@
...
+
+/// Get the latest installed NS Proton version
+pub fn get_local_ns_proton_version() -> Result<String, String> {
+    let compat_dir = get_proton_dir().unwrap();
+    let ns_prefix = "NorthstarProton-";
+    let pattern = format!("{}/{}*/version", compat_dir, ns_prefix);
+
+    let mut version: String = "".to_string();
+
+    for e in glob::glob(&pattern).expect("Failed to read glob pattern") {
```
GeckoEidechse, line 58: This iterates over all Proton folders, right?

GeckoEidechse (changes requested):
Some stuff I found from reading code and minor testing

src-tauri/src/platform_specific/linux.rs
Jan200101: going further, this should probably be made async (I think they are called tasks in tauri?) so the UI keeps rendering while its happening.

src-tauri/src/platform_specific/linux.rs
Jan200101: according to the documentation its always suppose to return PathBuf but life is sometimes more complicated.

Would be worth looking into how its implemented in the case there are some edge cases (e.g. returning empty pathbuf, panicing etc.)

src-tauri/src/platform_specific/linux.rs
GeckoEidechse: Oh yeah, good call. Though it should be enough to just make the wrapper functions async ^^

src-tauri/src/platform_specific/linux.rs
Jan200101: This iterates over all installations that use the official NorthstarProton folder structure.

This might not be reliable(?) but its only 1 of 2 places we can check (the other being the version file).

src-tauri/src/platform_specific/linux.rs
Jan200101: okay, just checked up on it.

Its *very* naive.

On systems without a temp dir it just panics, on others it uses assumptions (e.g. /tmp on unix) or calls relevant C functions without any error checking (e.g. GetTempPath2W on windows)

But its offical so /shrug

src-tauri/src/main.rs
```diff
@@ -628,3 +631,30 @@ fn launch_northstar_steam(
...
+async fn uninstall_northstar_proton_wrapper() -> Result<(), String> {
+    #[cfg(target_os = "linux")]
+    return linux::uninstall_ns_proton();
+
+    #[cfg(target_os = "windows")]
+    Err("Not supported on Windows".to_string())
+}
+
+#[tauri::command]
+fn get_local_northstar_proton_wrapper_version() -> Result<String, String> {
```
GeckoEidechse, line 654: Why is this one not `async`? ^^

src-tauri/src/platform_specific/linux.rs
```diff
@@ -1,8 +1,80 @@
...
+    Some(compat_dir)
+}
+
+/// Downloads and installs NS proton
+/// Assumes Steam install
+pub fn install_ns_proton() -> Result<(), thermite::prelude::ThermiteError> {
+    // Get latest NorthstarProton release
+    let latest = thermite::core::latest_release()?;
+
+    let temp_dir = env::temp_dir();
```
GeckoEidechse, line 20: I wouldn't namespace (or whatever it's called) a function if it's just called once

```suggestion
    let temp_dir = std::env::temp_dir();
```

src-tauri/src/platform_specific/linux.rs
```diff
@@ -1,8 +1,80 @@
 // Linux specific code
 
 use regex::Regex;
+use std::env;
```
GeckoEidechse, line 4: I wouldn't namespace (or whatever it's called) a function if it's just called once
```suggestion
```

src-tauri/src/main.rs
```diff
@@ -628,3 +631,30 @@ fn launch_northstar_steam(
 
     retval
 }
+
+#[tauri::command]
+async fn install_northstar_proton_wrapper() -> Result<(), String> {
```
GeckoEidechse, line 636: Maybe add a doc comment for wrapper functions as well. So for example

```suggestion
/// On Linux attempts to install NorthstarProton
/// On Windows simply returns an error message
#[tauri::command]
async fn install_northstar_proton_wrapper() -> Result<(), String> {
```

GeckoEidechse (commented):
Sorry I didn't review earlier. Took so long that by now it got some merge conflicts... x_x

src-tauri/src/main.rs
Jan200101: The function is so short and simple that I think its not worth spinning up a thread for it.
It would only really run into problems, if something was completely exhausting the Disk IO, but even then it would be a drop of water in an ocean.

Jan200101:
Successful clippy checks without testing it prior :sunglasses: (I totally am not forgetting to run it)

catornot:
> Successful clippy checks without testing it prior 😎 (I totally am not forgetting to run it)

Why aren't you using a lsp ( rust analyzer ); You wouldn't have to run anything.

Jan200101:
> Why aren't you using a lsp ( rust analyzer ); You wouldn't have to run anything.

Not a fan of having to setup an lsp for everything I work for.
Don't need more than syntax highlighting.

src-tauri/src/platform_specific/linux.rs
```diff
@@ -3,6 +3,79 @@
...
+
+/// Downloads and installs NS proton
+/// Assumes Steam install
+pub fn install_ns_proton() -> Result<(), thermite::prelude::ThermiteError> {
+    // Get latest NorthstarProton release
+    let latest = thermite::core::latest_release()?;
+
+    let temp_dir = std::env::temp_dir();
+    let path = format!("{}/nsproton-{}.zip", temp_dir.display(), latest);
+    let archive = std::fs::File::open(path.clone()).unwrap();
```
GeckoEidechse, line 21: So trying to run _Install NSProton_ I get a panic here:

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/platform_specific/linux.rs:21:53
```

Jan200101: Yep, turned out I bunged up in the migration to the libthermite change.
It should be `File::create().unwrap_or(File::open())`

Jan200101:
Test locally and it """works""".

As mentioned on Discord, the libthermite api for proton stuff is very bare and dreadful to use (with an API break in a minor release that made this PR even more painful).
With a potential migration to Proton GE in the future (or forking it, or using Proton Experimental because apparently that works now) a more in-depth solution would be needed.

But since this is a dev menu button, this is good to go as is.

src-vue/src/views/DeveloperView.vue
```diff
@@ -308,6 +320,21 @@ export default defineComponent({
                     notification.close();
                 });
         },
+        async installNSProton() {
+            await invoke("install_northstar_proton_wrapper")
+                .then((message) => { showNotification(`Done`); })
+                .catch((error) => { showNotification(`Error`, error, "error"); })
+        },
```
GeckoEidechse, line 327: My internet connection is so slow right now that I cannot test installation atm cause the download takes forever. My takeaway from this however is that it might make sense to spawn an initial notification telling the user that installation has begun and then clearing it again afterwards on success/failure.

Example reference:
https://github.com/R2NorthstarTools/FlightCore/blob/main/src-vue/src/views/DeveloperView.vue#L255-L268


Optimally we should have a progress bar but that's kinda outside the scope of the PR and requires changes upstream.

Jan200101: Sure, I can add a little notification stating "Started Proton Install"

This is basically just a minimally viable product, in the future this will be fleshed out and have all the libthermite stuff removed.

src-tauri/src/platform_specific/linux.rs
```diff
@@ -3,6 +3,81 @@
...
+    let compat_dir = get_proton_dir().unwrap();
+    std::fs::create_dir_all(compat_dir.clone())?;
+
+    let finished = std::fs::File::open(path.clone())?;
+
+    // Extract to Proton dir
+    log::info!("NorthstarProton installation started");
+    thermite::core::install_ns_proton(&finished, compat_dir)?;
+    log::info!("NorthstarProton installation finished");
+    drop(finished);
```
GeckoEidechse, line 37: Is this necessary. Shouldn't the value get dropped anyway on function exit or am I missing something here? ^^

src-tauri/src/platform_specific/linux.rs
```diff
@@ -3,6 +3,81 @@
...
+    let temp_dir = std::env::temp_dir();
+    let path = format!("{}/nsproton-{}.tar.gz", temp_dir.display(), latest);
+    let archive = std::fs::File::create(path.clone())?;
+
+    // Download the latest Proton release
+    log::info!("NorthstarProton download started");
+    thermite::core::download_ns_proton(latest, archive)?;
+    log::info!("NorthstarProton download finished");
+
+    let compat_dir = get_proton_dir().unwrap();
```
GeckoEidechse, line 28: Maybe write the directory to log. At least in testing on my laptop I had no idea where it was being installed to. Wanted to check whether it picked up the Flatpak install correctly ^^"

src-tauri/src/platform_specific/linux.rs
```diff
@@ -3,6 +3,81 @@
...
+    // Get latest NorthstarProton release
+    let latest = thermite::core::latest_release()?;
+
+    let temp_dir = std::env::temp_dir();
+    let path = format!("{}/nsproton-{}.tar.gz", temp_dir.display(), latest);
+    let archive = std::fs::File::create(path.clone())?;
+
+    // Download the latest Proton release
+    log::info!("NorthstarProton download started");
+    thermite::core::download_ns_proton(latest, archive)?;
```
GeckoEidechse, line 25: Would also write the temp dir path to log

GeckoEidechse (commented):
Seems to work in testing in the sense that the install didn't panic and checking version afterwards printed a value.

Cannot really test if it is actually able to run Titanfall2 once the proton install is selected as I don't have a Linux install running Titanfall2. Well I guess it's still a dev feature after all so I'm fine with what it does so far.

Once the remaining comments + merge conflicts are addressed it should be good to merge :D

src-tauri/src/platform_specific/linux.rs
Jan200101: Good idea, will add.

src-tauri/src/platform_specific/linux.rs
Jan200101: :+1:

src-tauri/src/platform_specific/linux.rs
GeckoEidechse: Aight. I would've assumed it gets dropped on function exit but doesn't hurt to do it explicitly ig ¯\\\_(ツ)\_/¯

Jan200101: Which one?
It gets dropped once we leave the local function.
`thermite::core::install_ns_proton` only takes a reference and thus move semantics do not kick in, leaving us with a File instance that has the file we are about to delete still open.

Jan200101: I assume you mean our `install_ns_proton`.
Yeah, at the end of the function the variable would be dropped, but I would like to discourage using it by explicitly dropping it before removing the file, which should come at no additional performance cost.

GeckoEidechse (approved):
Confirmed working in testing in the sense that installing, checking install, and removing works.

Also Steam seems to be able to pick it up

image


Didn't test actually running Northstar with it.

PR #390, opened 2023-06-12
fix: Hide deprecated mods in vanilla mods list

### LINKED ISSUE #386: Deprecated mods are still shown (with warning) when not performing search
#382 fixed showing deprecated mods however it seems to only apply to search, i.e. the mod will still show up when scrolling through all of Thunderstore (not doing a search and showing deprecated mods is disabled in settings)

image

image

Alystrasz:
Deprecated mods are filtered out from vanilla (*i.e.* not the one used while searching) mods list.

image

Closes #386.


GeckoEidechse (approved):
Confirmed working in testing. Thanks for the PR :D

PR #391, opened 2023-06-12
fix: Display install messages even outside mod view

### LINKED ISSUE #384: TypeError: this.$ is not a function
image

To reproduce

1. Search for `HaloAnnouncer_HUDRevamp_Compatibility_Patch`
2. Click on _Install_
3. While it's downloading, add some other characters to search so that the mod is no longer shown
4. Wait
5. Error should pop up

Alystrasz:
Mod installation notification translation would fail when invoked outside of the triggering Vue component; this addresses the issue by capturing the translation method in a context, so it's available when notification is fired up.

---

###### Without this PR

image

###### With this PR

image

---

Closes #384.

### DIFF
--- a/src-vue/src/components/ThunderstoreModCard.vue
+++ b/src-vue/src/components/ThunderstoreModCard.vue
@@ -251,8 +251,11 @@ export default defineComponent({
                 this.isBeingInstalled = true;
             }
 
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
The fix is like some TypeScript black magic to me but it works in testing when using the steps described in #384 to attempt to reproduce the issue ¯\\\_(ツ)\_/¯

PR #392, opened 2023-06-19
fix: Pass launch options to Steam

Jan200101:
relying on a file to be parsed in time is unreliably, and if the file is open while we delete it we may run into locking problems downside is that this gives a fugly prompt

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -453,10 +453,6 @@ async fn get_available_northstar_versions() -> Result<Vec<NorthstarThunderstoreR
     Ok(releases)
 }
 
-// The remaining below was originally in `lib.rs`.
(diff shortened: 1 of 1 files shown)


Jan200101:
pretend I didn't forget to run clippy

catornot:
> pretend I didn't forget to run clippy

you can configure rust-analyzer to use clippy instead of check

Jan200101:
> you can configure rust-analyzer to use clippy instead of check

this would require me to even use rust-analyzer, I just wing it.

Jan200101:
Tested this myself on Fedora 38

GeckoEidechse:
So on my Steam install I cannot launch Northstar at all right now using `-northstar` at rn. No clue what's causing that. So I'm just gonna trust you and merge anyway ^^

PR #393, opened 2023-06-23
feat: Add button to disable ModSettings mod

GeckoEidechse:
In preparation for Northstar releasing merged Mod Settings mod.

image

### DIFF
--- a/src-vue/src/i18n/lang/en.json
+++ b/src-vue/src/i18n/lang/en.json
@@ -119,6 +119,8 @@
                 "warning": "This window contains various functionality to repair common issues with Northstar and FlightCore.",
                 "disable_all_but_core": "Disable all but core mods",
                 "disable_all_but_core_success": "Disabled all mods but core",
+                "disable_modsettings": "Disable ModSettings mod",
+                "disable_modsettings_success": "Disabled ModSettings mod",
                 "force_reinstall_ns": "Force reinstall Northstar",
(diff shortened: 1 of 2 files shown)


GeckoEidechse:
@EladNLG Hi :)

EladNLG:
hello yes what do youk want

catornot:
you should probably also add a button to disable all mods that depend on ModSettings.

GeckoEidechse:
> you should probably also add a button to disable all mods that depend on ModSettings.

Do we need one? Cause AFAIK the merged code mirrors ModSettings functions so one can simply drop ModSettings mod, no?

catornot:
> > you should probably also add a button to disable all mods that depend on ModSettings.
> 
> Do we need one? Cause AFAIK the merged code mirrors ModSettings functions so one can simply drop ModSettings mod, no?

I though it was different `¯\_(ツ)_/¯`

ASpoonPlaysGames:
> Do we need one? Cause AFAIK the merged code mirrors ModSettings functions so one can simply drop ModSettings mod, no?

Not necessarily, there are two cases where they will break:

If a mod relies on pre 2.0 modsettings (should be very rare now, and not a lot we can do about it)

If a mod uses dependency constants. this will break the mod in some way, depends how it is made. Some mods will prompt the user to download modsettings, which is going to be confusing since it wont work

Alystrasz (approved):
Confirmed working on Windows.

PR #394, opened 2023-06-24
chore: replace game_scanner usage with winreg

### LINKED ISSUE #101: Replace `game-scanner` library with simple call to Windows registry
With #99 we use [`game-scanner`](https://crates.io/crates/game-scanner) as a library to get Origin Titanfall2 install on Windows. However this library adds quite a bunch of dependencies increasing build time.

@mrsteyk suggested using [`winreg`](https://crates.io/crates/winreg) to query Windows registry directly instead.

Jan200101:
closes #101

Not compiled on Windows (I don't have a setup for that, so someone else needs to do it).

Error message was updated to reflect this works on Origin and the EA App (should the text be changed to only mention the EA App, since Origin is considered defunct?), though the function name was not updated.

Moved winapi and winreg dependencies into the windows only dependencies, since other systems don't make use of them.

### DIFF
--- a/src-tauri/src/platform_specific/windows.rs
+++ b/src-tauri/src/platform_specific/windows.rs
@@ -10,27 +10,24 @@ use crate::check_is_valid_game_path;
 pub fn origin_install_location_detection() -> Result<String, anyhow::Error> {
     #[cfg(target_os = "windows")]
     {
-        let error;
         let hklm = RegKey::predef(HKEY_LOCAL_MACHINE);
         match hklm.open_subkey("SOFTWARE\\Respawn\\Titanfall2") {
-            Ok(tf) => match cur_ver.get_value("Install Dir") {
-                Ok(install_dir) => match check_is_valid_game_path(install_dir) {
+            Ok(tf) => {
+                let game_path_str: String = tf.get_value("Install Dir")?;
+
+                match check_is_valid_game_path(&game_path_str) {
                     Ok(()) => {
-                        return Ok(install_dir.to_string());
+                        return Ok(game_path_str.to_string());
                     }
                     Err(err) => {
-                        error = err;
+                        log::warn!("{err}");
                     }
-                },
-                Err(err) => {
-                    error = err;
                 }
-            },
+            }
             Err(err) => {
-                error = err;
+                log::warn!("{err}");
             }
         }
-        log::warn!("{}", error);
     }
 
     Err(anyhow!("No Origin / EA App install path found"))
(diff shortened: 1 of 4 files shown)


GeckoEidechse:
Currently also don't have a working Windows setup rn to test. Will do in a day or two hopefully.

src-tauri/src/platform_specific/windows.rs
```diff
@@ -1,34 +1,34 @@
...
-                        Err(err) => {
-                            log::warn!("{}", err);
-                            continue; // Not a valid game path
-                        }
+    #[cfg(target_os = "windows")]
+    {
+        let hklm = RegKey::predef(HKEY_LOCAL_MACHINE);
+        match hklm.open_subkey("SOFTWARE\\Respawn\\Titanfall2") {
+            Ok(tf) => {
+                let game_path_str: String = tf.get_value("Install Dir")?;
```
Jan200101, line 16: the type that `get_value` returns depend on the type that is expected, so we cannot easily set the type of game_path_str to something like `Result<String>` without the function returning a `Result<Result<String>>`.

The curent way its done is a bit ugly but it does the job (plus what are the chances someone has the registry subkey but not the value :clueless:)

Alystrasz (approved):
FlightCore found the path of my Steam install through registry successfully.

GeckoEidechse:
> FlightCore found the path of my Steam install through registry successfully.

Given that this affects only EA App. We should still test on a setup that only has EA App and no Steam. In particular, detecting Steam install is done via `library.vdf` and not registry.

Jan200101:
I'll setup a clean VM and test a fresh install from both Origin and the EA App.

Alystrasz:
My bad!

Jan200101:
Fresh Install of Titanfall 2 from the EA App

image

I'll test Origin later today with [Fuck_off_EA_App](https://github.com/p0358/Fuck_off_EA_App)

GeckoEidechse:
> Fresh Install of Titanfall 2 from the EA App

You're awesome, thanks <3



> I'll test Origin later today with [Fuck_off_EA_App](https://github.com/p0358/Fuck_off_EA_App)

Wouldn't mind but honestly not necessary. They should behave the same and tbh I'm not gonna consider it any more. Compared to EA App Origin is a minority and those few that still use it can probably also figure out how to manually set a path should it somehow fail ^^

src-tauri/src/platform_specific/windows.rs
GeckoEidechse: I can probably figure out a way to improve this but given how many dependencies we can slash by removing game scanner I'm fine with merging as is ^^

Jan200101:
On Origin (with an incomplete install. Why doesn't the EA App have this feature? its really cool!)

image

PR #399, opened 2023-06-28
feat: Add warning for Snap installations

### LINKED ISSUE #398: Detect and warn about Snap installs
Pressure Vessel and Snap do not appear to like each other so NSProton is completely broken.

This can easily be fixed by just using the official steam.deb, but users may think using the Ubuntu Store is the right way to go.

With this I propose adding detection for snap installs and warning the user about possible problems (and maybe adding information on the wiki to educate the user)

Jan200101:
Needs testing

closes #398

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -62,6 +62,7 @@ zip-extract = "0.1.2"
 # open urls
 open = "3.2.0"
 semver = "1.0"
+dirs = "5"
 
 [target.'cfg(windows)'.dependencies]
 # Windows API stuff
--- a/src-tauri/src/northstar/install.rs
+++ b/src-tauri/src/northstar/install.rs
@@ -168,6 +168,22 @@ pub fn find_game_install_location() -> Result<GameInstall, String> {
     // Attempt parsing Steam library directly
     match steamlocate::SteamDir::locate() {
         Some(mut steamdir) => {
+            #[cfg(target_os = "linux")]
+            {
+                let snap_dir = match std::env::var("SNAP_USER_DATA") {
+                    Ok(snap_dir) => std::path::PathBuf::from(snap_dir),
+                    Err(_) => match dirs::home_dir() {
+                        Some(path) => path,
+                        None => std::path::PathBuf::new(),
+                    }
+                    .join("snap"),
+                };
+
+                if steamdir.path.starts_with(snap_dir) {
+                    log::warn!("Found Steam installed via Snap, you may encounter issues");
+                }
+            }
+
             let titanfall2_steamid = TITANFALL2_STEAM_ID.parse().unwrap();
             match steamdir.app(&titanfall2_steamid) {
                 Some(app) => {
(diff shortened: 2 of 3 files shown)


src-tauri/src/northstar/install.rs
```diff
@@ -163,6 +163,22 @@ pub fn find_game_install_location() -> Result<GameInstall, String> {
     // Attempt parsing Steam library directly
     match steamlocate::SteamDir::locate() {
         Some(mut steamdir) => {
+            #[cfg(not(target_os = "linux"))]
```
Jan200101, line 166: I fixed this check up, but otherwise all is good.

src-tauri/Cargo.toml
```diff
@@ -68,6 +68,9 @@ open = "3.2.0"
 winapi = "0.3.9"
 semver = "1.0"
 
+[target.'cfg(not(target_os="windows"))'.dependencies]
```
Jan200101, line 71: should be updated to be linux, not not windows

Jan200101:
Rebasing to resolve conflicts and clean up the history.
No review existed and the prior state was broken so I force pushed.
Future changes will be additional commits.

src-tauri/src/northstar/install.rs
```diff
@@ -168,6 +168,22 @@ pub fn find_game_install_location() -> Result<GameInstall, String> {
...
+                    Ok(snap_dir) => std::path::PathBuf::from(snap_dir),
+                    Err(_) => match dirs::home_dir() {
+                        Some(path) => path,
+                        None => std::path::PathBuf::new(),
+                    }
+                    .join("snap"),
+                };
+
+                if steamdir.path.starts_with(snap_dir) {
+                    log::warn!("Found Steam installed via Snap, you may encounter issues");
```
GeckoEidechse, line 183: Just to be clear that this warning is not shown to end-user unless they run from CLI. I guess that's fine?

At least I don't know an easy way to propagate message back to UI. From the top of my head if we just return an `Error` it gets caugh somewhere higher up in the stack and just converted to a generic "Couldn't find install".

Might need to rework the error system to use custom error types but that's a whole other beast ^^"

Jan200101: I think returning an Error is a bit much.
This isn't a fatal error but known to cause problems.

Maybe the the dialogs from #418 would be good for this?
Open a dialog, ask if the user wants to launch via snap regardless of any issues and continue

GeckoEidechse: Yeah, could be an idea. I'd say we me merge as is, and then add dialog here once the other PR is merged.

Jan200101: Sounds good

GeckoEidechse:
As for testing, I guess we'll just do it in production ¯\\\_(ツ)\_/¯

Jan200101:
I've tested it by manually injecting values on an Ubuntu LTS Desktop install.

steamlocate-rs does not currently support snap, https://github.com/WilliamVenner/steamlocate-rs/pull/30

GeckoEidechse:
Welp, guess we just future proofed then a bit, haha

PR #401, opened 2023-06-29
feat: Delete old version when updating mod

### LINKED ISSUE #408: Delete old version of mod when updating
Related #401

Example https://northstar.thunderstore.io/package/DINOTICK/SpinnyAlternator/

Downloading `1.0.0` and then updating to `2.0.0` doesn't remove `1.0.0` causing a double install and a FlightCore that claims mod can be updated.

The solution is to nuke `1.0.0` when `2.0.0` installs succesfully.

GeckoEidechse:
Deletes all Northstar mods based on Thunderstore mod string before installing new version. This way we can clean up old files from mods before installing new version.

Note that the deletion is purposefully only done right before mod is extracted from install. This means it has already been downloaded. This is done to avoid the case where due to a slow internet connection an install fails but the old version of the mod is already removed due to the new logic.

Closes #408

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -435,6 +435,18 @@ pub async fn fc_download_mod_and_install(
     // Get Thunderstore mod author
     let author = thunderstore_mod_string.split('-').next().unwrap();
 
+    // Delete previous version of mod if already installed
+    // This way we ensure that any old files/folders are removed
+    match delete_thunderstore_mod(game_install.clone(), thunderstore_mod_string.to_string()) {
+        Ok(()) => (),
+        Err(err) => {
+            // Error can include the mod being found installed cause this is the first time it's being installed
+            // As such we just ignore any error (but still log it just in case)
+            log::info!("{err}");
+            ()
+        }
+    };
+
     // Extract the mod to the mods directory
     match thermite::core::manage::install_mod(
         author,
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Right now there's no way to tell in this part of the code whether we are dealing with an update or an install. Maybe we should refactor code to pass an argument that tells whether we are installing or updating?

GeckoEidechse:
(Haven't actually tested yet)

GeckoEidechse:
Tested with https://northstar.thunderstore.io/package/Creamy_jpg/Customisable_Reticles_Framework/ which seems to work. 

Although I need to check what happens if the update to install is incorrectly formatted. Pretty sure it just nukes the mod then but fails to install the new one...

GeckoEidechse:
> Although I need to check what happens if the update to install is incorrectly formatted. Pretty sure it just nukes the mod then but fails to install the new one...

Yup just tested with https://northstar.thunderstore.io/package/ScureX/ModCreator/ by manually putting it into mod folders, setting it up to look like older version and then trying to "update" it.

The result is that the old version gets remove but the new one fails to install which shouldn't happen IMO.

If the update fails, it should just keep the old version installed. Not sure what the best approach is here to achieve that... Any help would be appreciated <3

GeckoEidechse:
With https://github.com/R2Northstar/NorthstarLauncher/pull/503 this PR in its current state won't be necessary anymore

GeckoEidechse:
Closed by #415

PR #403, opened 2023-07-02
fix: delay scrolling the Mods View to prevent it stopping prematurely

Jan200101:
With the current logic the modview would change while its trying to scroll, resulting in the scrollbar getting messed up and the scrolling stopping in the middle of the page.

This PR:
- Uses `update:current-page` event for setting the page index
- Uses `current-change` for scrolling
(`update:current-page` is emitted before `current-change`)
- delays the `scrollTo` call by 100ms


Alystrasz (approved):
I could reproduce the issue on Ubuntu 22.04, but not with Windows 10?
Anyway, this fixes mentioned issue, so fine by me.

PR #409, opened 2023-07-07
chore: Bump libthermite to `0.6.5`

### LINKED ISSUE #406: HUD Revamp 2.0.0 fails to install
image

Technically a libthermite issue but given that we are a few versions behind, I'm gonna keep the issue here for now.

GeckoEidechse:
Bumps `libthermite` to newest version which should hopefully fix #406 

The library had some API changes from `0.5` to `0.6` that need to be accounted for.

Closes #406


GeckoEidechse:
Aight, seems to work in testing so far. Installed some mods and Northstar ^^

Issue #415, opened 2023-07-16
Support new `packages` mod format


GeckoEidechse:
Related to:
- https://github.com/R2Northstar/NorthstarLauncher/pull/503
- https://github.com/R2Northstar/NorthstarLauncher/issues/505

Supersedes:
- https://github.com/R2NorthstarTools/FlightCore/issues/288

GeckoEidechse:
Support is basically complete with `v2.0.0`. The only thing missing is a way for players to upgrade their existing mods to packages (where applicable). Right now this only happens on reinstalling or updating a mod.

Preferably we want to accelerate that process by e.g. prompting player to "upgrade".

GeckoEidechse:
Apart from a way to get users to upgrade mods to packages this is now implemented with `v2.0.0`.

PR #417, opened 2023-07-16
feat: Initial support for reading NS packages dir

GeckoEidechse:
TODO:

- [ ] Deleting Northstar mod does not clean up entire packages dir after deletion
- [ ] Code cleanup
- [ ] ???

Part of #415


Alystrasz:
A few questions, for context, before I review:
* I suppose those FlightCore PRs follow https://github.com/R2Northstar/NorthstarLauncher/pull/503, right?
* Why do I feel like the code was copy-pasted from an already-existing module, while (in my mind) reading mods from a new directory should "simply" be done by adding a reference to previously-mentioned directory?

Alystrasz (approved):
Code is good and works on my Windows 10 machine.

image

PR #418, opened 2023-07-17
feat: Replace MessageBoxW invocation with Tauri dialog

Jan200101:
With Tauri 2.0 this will require a plugin

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -23,7 +23,7 @@ tauri-build = { version = "1.4", features = [] }
 [dependencies]
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
-tauri = { version = "1.4", features = ["api-all", "updater"] }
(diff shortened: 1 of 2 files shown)


Jan200101:
Copied the textbox code into `linux_checks` for testing

Here it is on Linux:
image

Will test on Windows later

Jan200101:
Tested on Windows using the same method as before

image

Pressing Ok opens the link, closing the window just closes it.

Jan200101:
Its worth noting that this is non blocking and idk if this will cause problems when the main window wants to exit

GeckoEidechse:
> It's worth noting that this is non blocking and idk if this will cause problems when the main window wants to exit

Don't have access to a Windows machine rn, you should be able to test that by just uninstalling WebView2 and then trying to run FlightCore and check if the behaviour is the same as before ^^

Jan200101:
Tested and it did indeed break.

The blocking version does the job though.
image

GeckoEidechse (approved):
Works as expected in testing on Windows with WebView2 not installed:

image

PR #419, opened 2023-07-17
Add support for launch parameters

Jan200101:


src-tauri/src/main.rs
```diff
@@ -474,6 +474,7 @@ pub enum InstallType {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct GameInstall {
     pub game_path: String,
+    pub launch_parameters: String,
```
Jan200101, line 477: adding it to GameInstall is not ideal, but the alternative is passing it to the function directly, which would change its signature.
Consider this a stop-gap implementation

GeckoEidechse: You can make function signature use `Option<String>` for launch args. That changes the signature on Rust side but doesn't require any change on UI side, as missing function arg on ui is counted as `None` when sent to backend ^^

Jan200101: Yeah, but its doesn't fit logically.
Its not part of a "game install".

I think it would fit better to pass the launch options directly to Rust, either via a string or an array (a singular string would cause problems with flags that have quoted values)

Alystrasz:
What's the difference with #301?
No anger here, I just don't wanna spoil our development efforts :)

Jan200101:
> What's the difference with #301? No anger here, I just don't wanna spoil our development efforts :)

Oh, I hadn't seen that PR.
Overall I would say
- variable parameters
- are directly passed to NS instead of using ns_startup_args.txt, which would persist beyond the lifetime of FlightCore

A combination of the two with a text input that lets you add your own parameters while giving you a selection of known commands would probably be best.

Having already tried the file based approach with #178 I think it would be best to pass the options directly, since those files persist outside of FlightCore and you cannot control what other stuff may be passed.

I'm open for going down whatever route, I'd just love to be able to use FlightCore for all thats needed to launch Northstar.

GeckoEidechse:
I'm not sure how I feel about storing launch args in FlightCore persistent store (only now got the chance to read the full PR)...

For background when I initially set out on making FlightCore I didn't want any persistence at all. But alas people have weird installs so I need to keep track of them when I cannot autodetect. Also having dev stay enabled is kinda nice, if you wanna change the UI language, you don't wanna do it every time, etc. so I was kinda forced to use persistence after all.

The way I see it, I would keep this PR to just adding the option to launch with additional launch args. So that for example one could then easily add a button that e.g. launches a custom profile.

The way I see it, the `ns_launch_args.txt` is preferred for args that are across multiple profiles, e.g. `-multiple` or `+exec custom.cfg` while having the option to allow for additional args is useful when tying it to a profile.

All that aside, I would keep this PR to just adding support for launch args but without the store part. With persistence, any additional thing we store increases the number of combinations exponentially, which makes it easy to introduce bugs. So I'd say either remove it from this PR or make a separate one without it and then we merge the non-persistence stuff first.

If you wanna have a way to test launch args, you could add a button that has some hardcoded to e.g. launch the PR profile ;)

src-tauri/src/main.rs
Jan200101: Ah, misunderstood you there.

PR #422, opened 2023-07-17
feat: Launch via Steam on Non Windows Systems if possible

Jan200101:
Launching via Steam has been in the dev menu for a hot minute so I think its worth wiring it up to the Launch Game button when not on Windows and installed via Steam.

Tested on my Fedora 39 KDE instance.

### DIFF
--- a/src-tauri/src/northstar/mod.rs
+++ b/src-tauri/src/northstar/mod.rs
@@ -3,7 +3,7 @@
 pub mod install;
 
 use crate::util::check_ea_app_or_origin_running;
-use crate::{constants::CORE_MODS, get_host_os, GameInstall, InstallType};
+use crate::{constants::{CORE_MODS, TITANFALL2_STEAM_ID}, get_host_os, GameInstall, InstallType};
 use anyhow::anyhow;
 
 /// Check version number of a mod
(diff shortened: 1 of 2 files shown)


src-tauri/src/northstar/mod.rs
```diff
@@ -65,16 +68,16 @@ pub fn launch_northstar(
...
-            || matches!(game_install.install_type, InstallType::ORIGIN)
-            || matches!(game_install.install_type, InstallType::UNKNOWN))
-    {
-        return Err(format!(
-            "Not yet implemented for \"{}\" with Titanfall2 installed via \"{:?}\"",
-            get_host_os(),
-            game_install.install_type
-        ));
+    if host_os != "windows" {
+        if !matches!(game_install.install_type, InstallType::STEAM) {
```
GeckoEidechse, line 72: Note to self: While this technically fucks with the logic so that now `InstallType::EAPLAY` is now no longer caught as not launchable, this is fine as:

1. EA App is not properly detected anyway as it's caught by the same detection mechanism as Origin and as such  is always misclassified as "Origin"
2. Launching EA App works just fine as long as the location is read/write which is caught by an earlier check during Northstar installation.

src-tauri/src/northstar/mod.rs
```diff
@@ -122,3 +125,56 @@ pub fn launch_northstar(
...
+            if get_host_os() != "windows" {
+                let titanfall2_steamid: u32 = TITANFALL2_STEAM_ID.parse().unwrap();
+                match steamdir.compat_tool(&titanfall2_steamid) {
+                    Some(compat) => {
+                        if !compat
+                            .name
+                            .clone()
+                            .unwrap()
+                            .to_ascii_lowercase()
+                            .contains("northstarproton")
```
GeckoEidechse, line 150: I assume that check will be expanded later with Proton-GE?
(can be done in separate PR)

GeckoEidechse (approved):
I don't really have a Linux setup with Titanfall2 to test, so I'm just gonna trust it works as it did for you ^^

src-tauri/src/northstar/mod.rs
Jan200101: Depends on how things will be going forward.
I recall the plan being to modify the GE-Proton 8-4 release, add the wsock override and some more.

Idk if Tony is intended to rebrand it as NorthstarProton or not, we'll have to see.

GeckoEidechse: Aight, for now we'll just keep it as is I guess ^^

PR #423, opened 2023-07-17
feat: Compare Flightcore versions as semver

Jan200101:
Worked on PRs before the latest release was tagged causing a notification saying "Update Flightcore, you are running {newer version}. Newest ist {older version}"

### DIFF
--- a/src-tauri/src/github/release_notes.rs
+++ b/src-tauri/src/github/release_notes.rs
@@ -1,4 +1,5 @@
 use crate::constants::APP_USER_AGENT;
+use semver::Version;
 use serde::{Deserialize, Serialize};
 use std::vec::Vec;
 use ts_rs::TS;
@@ -58,14 +59,13 @@ pub async fn get_newest_flightcore_version() -> Result<FlightCoreVersion, String
 #[tauri::command]
 pub async fn check_is_flightcore_outdated() -> Result<bool, String> {
     let newest_flightcore_release = get_newest_flightcore_version().await?;
+    let newest_version = Version::parse(&newest_flightcore_release.tag_name[1..]).unwrap();
 
-    // Get version of installed FlightCore...
-    let version = env!("CARGO_PKG_VERSION");
-    // ...and format it
-    let version = format!("v{}", version);
+    // Get version of installed FlightCore
+    let current_version = env!("CARGO_PKG_VERSION");
+    let current_version = Version::parse(current_version).unwrap();
 
-    // TODO: This shouldn't be a string compare but promper semver compare
-    let is_outdated = version != newest_flightcore_release.tag_name;
+    let is_outdated = current_version < newest_version;
 
     // If outdated, check how new the update is
     if is_outdated {
--- a/src-tauri/src/github/release_notes.rs
+++ b/src-tauri/src/github/release_notes.rs
@@ -1,5 +1,4 @@
 use crate::constants::APP_USER_AGENT;
-use semver::Version;
 use serde::{Deserialize, Serialize};
 use std::vec::Vec;
 use ts_rs::TS;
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Oh fuck, I should've updated the comment. The non-semver compare is kinda on purpose cause it means that I can rollback a release by simply deleting it off of GitHub as opposed to making a new tag which tends to take like half an hour due to slow compile times in CI.

On the flipside, getting the notification every time while making a release is also kinda annoying, haha.

idk, I think it's best to just stay with the non-semver one for easy rollback and just update the comment ^^"

Jan200101:
Hmm... I'll see if there is an alternative solution to this...

GeckoEidechse:
I'm gonna set this to draft in the meantime to better see which PR need to be reviewed atm ^^

Jan200101:
Alternative Idea:
disabling the update check for dev builds.
This would be as simple as adding 
```

**Comment by GeckoEidechse** [comment]
Hmm, that could work, although I admittedly like annoying everyone into keeping their PRs up to date :>

IMO in terms of usage the ideal approach would be to do semver check on dev and non-semver in prod. That way I can still annoy everyone to update their PR but you won't get a warning if you update early.

Whether that's actually worth implementing, idk ^^"

**Comment by Jan200101** [comment]
Okay, then lets take that approach

Keep the semver code in this PR but:
- in dev `outdated = cur_ver < newest_version`
- else `outdated = cur_ver != newest_version`

This would keep the current behavior while also making too fresh builds not cause a notification.

**Comment by GeckoEidechse** [comment]
Sounds good to me :D

File: src-tauri/src/github/release_notes.rs [hunk 1]
```diff
@@ -58,14 +59,13 @@ pub async fn get_newest_flightcore_version() -> Result<FlightCoreVersion, String
 #[tauri::command]
 pub async fn check_is_flightcore_outdated() -> Result<bool, String> {
     let newest_flightcore_release = get_newest_flightcore_version().await?;
+    let newest_version = Version::parse(&newest_flightcore_release.tag_name[1..]).unwrap();
```
Jan200101, line 62: Would it make sense to have a default value here using `unwrap_or(_else)` ?

The function is async and will run in a thread, so any errors would be non fatal, but this could at least be useful when `CARGO_PKG_VERSION` is not proper semver.

src-tauri/src/github/release_notes.rs
```diff
@@ -58,14 +59,13 @@ pub async fn get_newest_flightcore_version() -> Result<FlightCoreVersion, String
...
 
-    // Get version of installed FlightCore...
-    let version = env!("CARGO_PKG_VERSION");
-    // ...and format it
-    let version = format!("v{}", version);
+    // Get version of installed FlightCore
+    let current_version = env!("CARGO_PKG_VERSION");
+    let current_version = Version::parse(current_version).unwrap();
```
Jan200101, line 66: same here

src-tauri/src/github/release_notes.rs
GeckoEidechse: What would the default value be? `0.0.0`?

I think as it's non fatal it's fine to just be an unwrap. In fact, if it fails I will get notified via crash handler and can address accordingly ^^

Jan200101: Hmmm, I think it would just be better to let it go to the crash handler instead.

Jan200101:
Tested by manually bumping `version` to `1.25.0` and then running with `npx tauri dev` for dev mode and `npx tauri dev --release` for release mode

GeckoEidechse (approved):
Seems good in quick testing on my end when bumping/lowering version numbers in Cargo.toml

PR #425, opened 2023-07-17
feat: Add ability to delete package

GeckoEidechse:
- Depends on #417

Part of #415


GeckoEidechse (commented):
Confirmed working in testing on Windows 10

PR #426, opened 2023-07-18
feat: Support installing Thunderstore packages

GeckoEidechse:
- [x] Best to have #417 first cause otherwise installed mods cannot be read.
- [x] Safety check that mod is actually correctly formatted
- [x] Uninstall older version of package if installed
- [x] ~~Wait for libthermite stabilisation~~ Current version seems stable enough

Part of #415

### DIFF
--- a/src-tauri/src/mod_management/legacy.rs
+++ b/src-tauri/src/mod_management/legacy.rs
@@ -117,6 +117,10 @@ pub fn parse_installed_mods(
 
 /// Deletes all legacy packages that match in author and mod name
 /// regardless of version
+/// 
+/// "legacy package" refers to a Thunderstore package installed into the `mods` folder
+/// by extracting Northstar mods contained inside and then adding `manifest.json` and `thunderstore_author.txt`
+/// to indicate which Thunderstore package they are part of
 pub fn delete_legacy_package_install(
     thunderstore_mod_string: &str,
     game_install: &GameInstall,
--- a/src-tauri/src/mod_management/legacy.rs
+++ b/src-tauri/src/mod_management/legacy.rs
@@ -115,6 +115,42 @@ pub fn parse_installed_mods(
     Ok(mods)
 }
 
+/// Deletes all legacy packages that match in author and mod name
+/// regardless of version
+pub fn delete_legacy_package_install(
+    thunderstore_mod_string: &str,
+    game_install: &GameInstall,
+) -> Result<(), String> {
+    let thunderstore_mod_string: ParsedThunderstoreModString =
+        thunderstore_mod_string.parse().unwrap();
+    let found_installed_legacy_mods = match parse_installed_mods(game_install) {
+        Ok(res) => res,
+        Err(err) => return Err(err.to_string()),
+    };
+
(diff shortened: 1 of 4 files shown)


GeckoEidechse:
For testing I used
- `Alystrasz-Parkour-0.1.1` as a mod that should pass instal
- `ScureX-ModCreator-1.0.0` as a mod that should fail install

GeckoEidechse:
Also tested updating an existing mod where a newer version is broken. It correctly didn't install new version and it kept old version.

GeckoEidechse:
CI error is due to clippy due to libthermite marking certain functions as deprecated early. Will look into how to best address but shouldn't prevent actually testing ^^"

src-tauri/src/mod_management/legacy.rs
```diff
@@ -115,6 +115,42 @@ pub fn parse_installed_mods(
     Ok(mods)
 }
 
+/// Deletes all legacy packages that match in author and mod name
```
Alystrasz, line 118: Could you be more precise regarding to what is a legacy package?

src-tauri/src/mod_management/mod.rs
```diff
@@ -407,6 +408,93 @@ async fn get_mod_dependencies(thunderstore_mod_string: &str) -> Result<Vec<Strin
...
+        }
+    }
+
+    for directory in directories {
+        let folder_name = directory.file_name().unwrap().to_str().unwrap();
+        let ts_mod_string_from_folder: ParsedThunderstoreModString = match folder_name.parse() {
+            Ok(res) => res,
+            Err(err) => {
+                log::warn!("{err}");
+                continue;
```
Alystrasz, line 448: Why continuing if there was an error?

src-tauri/src/mod_management/legacy.rs
GeckoEidechse: Sure. "Legacy package" is what I call a Thunderstore package installed into `mods` the old-fashioned way, i.e. how we did it up to now as opposed to using the `packages` folder.

Should I update the comment with something like that? ^^

Alystrasz: Yep, please add this to the documentation :)

src-tauri/src/mod_management/mod.rs
Alystrasz: Please write this in the documentation as well :)

Alystrasz (approved):
Omitting missing documentation, the feature works as expected (both in installing correcly-formatted mods and preventing install badly-formatted ones) on Windows 10 and Ubuntu 23.04.

src-tauri/src/mod_management/legacy.rs
GeckoEidechse: Done with 448c6f343984b48b3ba7e1b9faaab54651b758f9

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Done with e4f995b1f8e25ed6cddae2672ea3e79da80e55ee

PR #427, opened 2023-07-18
feat: Mechanism to upgrade to packages

GeckoEidechse:
We need some way for user to change their setup to `packages` I'm too time limited to get a proper UI and everything so I propose we hijack our existing update prompt for mods to achieve this.

Simply put:

- Give an outdated version number on legacy installed Thunderstore mods (I'm using `0.0.0` for this which is not available on Thunderstore)
- User has the option to "update mod"
- Adjust our update mechanism so that it also removes legacy versions of installed mods on successful update.

### DIFF
--- a/src-tauri/src/mod_management/legacy.rs
+++ b/src-tauri/src/mod_management/legacy.rs
@@ -88,7 +88,7 @@ pub fn parse_installed_mods(
             }
         };
         // Get Thunderstore mod string if it exists
-        let thunderstore_mod_string = match parsed_mod_json.thunderstore_mod_string {
+        let mut thunderstore_mod_string = match parsed_mod_json.thunderstore_mod_string {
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
There might be a better mechanism to do this, other than pretending mod is outdated.

Also I'm pretty sure having multiple mods outdated and spam clicking them to update causes some issues (at least based on past experience where install would then fail sometimes). So if we go with this approach we might wanna have an "update all" option or a way to block installing/updating another mod while current one is being installed.

Alystrasz:
> Give an outdated version number on legacy installed Thunderstore mods (I'm using 0.0.0 for this which is not available on Thunderstore)

If it works, it ain't stupid, right?

> or a way to block installing/updating another mod while current one is being installed.

Another issue for another day 😄

PR #429, opened 2023-07-18
dev: Packages

GeckoEidechse:
This PR is just to see the diff between all missing `packages` related changes and `main`

It contains
- #417 
- #425 
- #426 
- #427 
- As well as a few changes not yet PR'd

As such this PR should stay draft and not be merged on its own but rather to be used as a PoC and see overall progress of implementing the `packages` format.


GeckoEidechse:
So this is working now in testing. The code is a bit messy and partially duplicated but as I mentioned before I'm time limited and this works fine for the release. Once players start transitioning over to packages we can clean up a bunch of things 👀

GeckoEidechse:
Also in the "upgrade to packages" part I should probably also add a sort of exception for mod settings where in the upgrade process mod settings gets skipped so it just gets deleted instead as it's not in the game anyway.

GeckoEidechse:
Also I'd say the release that has everything packages related merged should be a 2.0 ^^

src-tauri/src/mod_management/mod.rs
```diff
@@ -407,6 +408,93 @@ async fn get_mod_dependencies(thunderstore_mod_string: &str) -> Result<Vec<Strin
...
+            has_mods = true;
+            if let Some(name) = file_path.file_name() {
+                if name == "mod.json" {
+                    let parent_path = file_path.parent().unwrap();
+                    if parent_path.parent().unwrap().to_str().unwrap() == "mods" {
+                        mod_json_exists = true;
+                    }
+                }
+            }
+        }
```
catornot, line 492: this is literately the perfect use for a iterator.
no mutability :)

src-tauri/src/mod_management/mod.rs
```diff
@@ -407,6 +408,93 @@ async fn get_mod_dependencies(thunderstore_mod_string: &str) -> Result<Vec<Strin
...
+            }
+        };
+        // Check which match `AUTHOR-MOD` and do NOT match `AUTHOR-MOD-VERSION`
+        if ts_mod_string_from_folder.author_name == thunderstore_mod_string.author_name
+            && ts_mod_string_from_folder.mod_name == thunderstore_mod_string.mod_name
+            && ts_mod_string_from_folder.version != thunderstore_mod_string.version
+        {
+            delete_package_folder(&directory.display().to_string())?;
+        }
+    }
```
catornot, line 458: a iterator would also be better here although you probably keep it and just move all these if cases into filters.

src-tauri/src/mod_management/legacy.rs
```diff
@@ -115,6 +126,42 @@ pub fn parse_installed_mods(
...
+            .parse()
+            .unwrap();
+
+        if thunderstore_mod_string.author_name == current_mod_ts_string.author_name
+            && thunderstore_mod_string.mod_name == current_mod_ts_string.mod_name
+        {
+            // They match, delete
+            delete_mod_folder(&legacy_mod.directory)?;
+        }
+    }
```
catornot, line 160: > a iterator would also be better here although you probably keep it and just move all these if cases into filters.

src-tauri/src/mod_management/mod.rs
```diff
@@ -486,19 +573,45 @@ pub async fn fc_download_mod_and_install(
...
+                    "Mod failed sanity check during install. It's probably not correctly formatted"
+                        .to_string(),
+                ),
+                _ => Err(err.to_string()),
+            };
+        }
+    };
+
+    // Successful package install
+    match legacy::delete_legacy_package_install(thunderstore_mod_string, game_install) {
```
catornot, line 602: consider using a `if let` here

src-tauri/src/mod_management/mod.rs
```diff
@@ -486,19 +573,45 @@ pub async fn fc_download_mod_and_install(
...
+        Ok(()) => (),
+        Err(err) => {
+            // Catch error but ignore
+            log::warn!("Failed deleting legacy versions due to: {}", err);
+        }
+    };
+
+    match delete_older_versions(thunderstore_mod_string, game_install) {
```
catornot, line 610: and here

src-tauri/src/mod_management/mod.rs
```diff
@@ -407,6 +408,93 @@ async fn get_mod_dependencies(thunderstore_mod_string: &str) -> Result<Vec<Strin
...
+
+    // Get list of folders in `mods` directory
+    for path in paths {
+        let my_path = path.unwrap().path();
+
+        let md = std::fs::metadata(my_path.clone()).unwrap();
+        if md.is_dir() {
+            directories.push(my_path);
+        }
+    }
```
catornot, line 440: for loop found, use a iterator instead in this case. ( I hate for loops ).

src-tauri/src/mod_management/legacy.rs
```diff
@@ -100,6 +100,17 @@ pub fn parse_installed_mods(
...
+        // This is a stupid way to show a legacy installed mod as outdated by simply giving back a wrong version number
+        if thunderstore_mod_string.is_some() {
+            // Parse the string
+            let mut parsed_string: ParsedThunderstoreModString =
+                thunderstore_mod_string.clone().unwrap().parse().unwrap();
+            // Set version number to `0.0.0`
+            parsed_string.version = "0.0.0".to_string();
+            // And store new string back in original variable
+            thunderstore_mod_string = Some(parsed_string.to_string())
+        }
```
catornot, line 112: perhaps just move this into the match case above ;)

catornot (changes requested):
I just like to complain about rust code.
also pls pls just actually use the result enum instead of unwrapping everything.

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Uh, I'm dumb, can I have an example? ^^

src-tauri/src/mod_management/mod.rs
GeckoEidechse: Yeah so I kinda merged #426 before seeing your comments on here. Will do some refactoring to change update to your feedback though ^^

src-tauri/src/mod_management/legacy.rs
GeckoEidechse: Wouldn't I have to duplicate the code then? Cause the `match` matches for like two cases
1. checking if it's in `mods.json`
2. checking if it there's a `manifest.json` and a `thunderstore_author.txt`

src-tauri/src/mod_management/mod.rs
catornot: smth like this 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=93be605e794e1939f0deeaeebb767575

it might not compile but it shouldn't be hard to fix stuff in it I think.

PR #432, opened 2023-07-19
i18n: Enable Simplified Chinese translations

GeckoEidechse:
Adds _Simplified Chinese_ to the language dropdown menu in settings.

image

### DIFF
--- a/src-vue/src/components/LanguageSelector.vue
+++ b/src-vue/src/components/LanguageSelector.vue
@@ -46,6 +46,10 @@ export default defineComponent({
                 value: 'it',
                 label: 'Italiano'
             },
+            {
+                value: 'zh_Hans',
+                label: '简体中文'
+            },
         ]
     }),
     mounted: async function () {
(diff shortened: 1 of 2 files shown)


src-vue/src/components/LanguageSelector.vue
```diff
@@ -46,6 +46,10 @@ export default defineComponent({
                 value: 'it',
                 label: 'Italiano'
             },
+            {
+                value: 'zh_Hans',
+                label: '简体中文'
```
GeckoEidechse, line 51: I don't speak Chinese so I used Google Translate for that key. Maybe a native speaker like @zxcPandora or @wolf109909 could confirm whether it's correct ^^"

GeckoEidechse: Seems to be fine: https://en.wiktionary.org/wiki/%E7%AE%80%E4%BD%93%E4%B8%AD%E6%96%87

PR #435, opened 2023-07-19
feat: Update to new wallpaper

GeckoEidechse:
I'm thinking if I already bump FlightCore version to 2.0 for `packages` support, why not also update the wallpaper for a change? :>

image

<details>
<summary>Old image</summary>

image

</details>


GeckoEidechse:
Draft for now so it doesn't accidentally get merged until everything else is ready ^^

GeckoEidechse:
@Alystrasz thoughts on the wallpaper? 👀

Alystrasz:
Not so sure about that wallpaper 😁 

* Style of the image is not unified, *e.g.* I think the Northstar 3D model pops out too much regarding the background;
* There shouldn't be any image elements under the menu bar area, the image part there should be as uniform as possible to ease reading menu items.

I suggest looking into the numerous Titanfall concept arts 😃

GeckoEidechse:
> * There shouldn't be any image elements under the menu bar area, the image part there should be as uniform as possible to ease reading menu items.

Agreed. I might be able to tweak the image a bit

> I suggest looking into the numerous Titanfall concept arts :smiley:

Ye, I primarily was thinking something Northstar related and something where the image isn't super noisy. For example I was considering using the one from Northstar



but decided against it due to it just being too "noisy" ^^

Jan200101:
I agree that the above image is too noisy, there is way too much things that draw your attention away.

At the same time, the proposed image feels empty with the FlightCore pose and absolutely nothing else.

Maybe some minor effects should be re-added, like the rocket thrusters and the lens flare.

GeckoEidechse:
I updated it :)

image

Jan200101 (approved):
I like it :+1:

Alystrasz (approved):
Submenus are still readably, it's a yes from me.

image

Alystrasz:
@GeckoEidechse maybe add the image credits in the About section :)

GeckoEidechse:
> @GeckoEidechse maybe add the image credits in the About section :)

Do I need to if the image is my own? ^^"

Alystrasz:
> > @GeckoEidechse maybe add the image credits in the About section :)
> 
> Do I need to if the image is my own? ^^"

Let's say we'll be fine!

PR #437, opened 2023-07-21
feat: Bring back the coloured menu bar

GeckoEidechse:
Adds the the background colour and blur back to the menu bar.

Looks like:
image


GeckoEidechse:
As per @EM4Volts request

EM4Volts:
sad, @Alystrasz i really liked the feature, it gives the ui a better overall look, makes the menu buttons stand out instead of being just there.

PR #438, opened 2023-07-21
feat: Prevent mods with plugins to be installed

GeckoEidechse:
This is an initial implementation to prevent mods with plugins from being installed.

The goal is to get something like #267 where plugins can be installed if the user opts into installing plugins. Maybe with a warning if they wanna continue etc.


GeckoEidechse (commented):
Prevents plugin install in testing. Again this is just temporary and we should soon add a way to allow plugin install, maybe copying some things from #267

PR #439, opened 2023-07-21
chore: Bump Rust dependencies

GeckoEidechse:
Haven't bumped Rust dependencies in a while so this is gonna be fun :>


GeckoEidechse (commented):
Seems to work fine in quick testing on Linux

PR #441, opened 2023-07-23
feat: Add basic warning on plugin install

GeckoEidechse:
Plugins in Northstar have unrestricted device access. As such we want to warn user before installing one with option to abort.

This implementation does not support localisation and the like as it uses Tauri dialogs #418 which are not translated via `vue-i18n`. 
The long term goal is to have the prompt in frontend to allow for localisation but doing it this way for now allow for rapid implementation and deployment ^^

image

For testing you can try to install [`cat_or_not-Furnace-1.0.0`](https://northstar.thunderstore.io/package/cat_or_not/Furnace/1.0.0)


GeckoEidechse:
@catornot :)

catornot (approved):
no complaints. :)

GeckoEidechse (commented):
Works in testing on Windows 10

PR #445, opened 2023-07-23
chore: Store `game_path` and `install_type` as `GameInstall` in store

Jan200101:
Only did very basic tests.

### DIFF
--- a/src-vue/src/plugins/store.ts
+++ b/src-vue/src/plugins/store.ts
@@ -212,7 +212,7 @@ export const store = createStore<FlightCoreStore>({
 
                 // Game is ready to play.
                 case NorthstarState.READY_TO_PLAY:
-                    await invoke("launch_northstar", { gameInstall: game_install })
+                    await invoke("launch_northstar", { gameInstall: state.game_install })
                         .then((message) => {
                             console.log(message);
                             // NorthstarState.RUNNING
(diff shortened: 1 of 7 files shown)


src-vue/src/plugins/store.ts
```diff
@@ -61,8 +60,7 @@ export const store = createStore<FlightCoreStore>({
     state(): FlightCoreStore {
         return {
             developer_mode: false,
-            game_path: undefined as unknown as string,
-            install_type: undefined as unknown as InstallType,
+            game_install: undefined as unknown as GameInstall,
```
Jan200101, line 63: We may want to use an object with default values to prevent any race condition during development.

GeckoEidechse: Given that it worked fine before moving the two values into an object I think it should be fine ^^

GeckoEidechse:
Being a meany and just pushed directly into your branch >:)

With that I cannot find any missing fields any more that still need updating. Anything from your side cause otherwise I'd say this should be good to merge ^^

src-vue/src/plugins/store.ts
Jan200101: Yeah but now we are trying to access members of `undefined`, resulting in this  error:
`Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating '_ctx.$store.state.game_install.game_path')`

Turning on my Web Dev brain for a moment, just using an empty object should be enough since all uninitialized members will become unknown.

PR #446, opened 2023-07-23
feat: Add button to DevView to calculate checksums

GeckoEidechse:
Adds button to DevView to calculate checksum of files in game install location. The results are then sent to text field in frontend from where it can be copy/pasted.

image

The idea behind this being to get contributors and/or playtesters to use it to crowdsource the checksums of various gamefiles to then roll our own game verification code as part of the repair window.

(Yes I'm aware that game files can be verified via Steam / EA App but that excludes files we add like `wsock32.dll` and I also want everything self contained so that in the long run I could add a "troubleshoot my game" button that runs functions for detecting the most common issues)

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -87,6 +87,7 @@ dependencies = [
  "async-recursion",
  "chrono",
  "const_format",
+ "crypto-hash",
  "dirs",
  "glob",
  "json5",
@@ -108,6 +109,7 @@ dependencies = [
  "tauri-plugin-store",
  "tokio",
  "ts-rs",
+ "walkdir",
  "winapi",
  "winreg 0.11.0",
  "zip",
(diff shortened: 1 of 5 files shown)


Alystrasz:
Could you describe in details the usecase for this feature again?
(First time I triggered the feature, FlightCore and vscode crash, cannot reproduce for now)

Maybe we can also use a notification displaying some progress, instead of a "Please wait" message :)

GeckoEidechse:
> Could you describe in details the usecase for this feature again?

Plan is to have a bunch of people that know how to access DevView run this on their machine so that we can grab a bunch of checksums from the game files. We need multiple to compare in case someone has some individual local changes (like a vpk mod for example).

After that, I can add the matching checksums to FlightCore and have some function that basically checks that all files match and throws an error for any file that doesn't match basically adding a file verification system ^^

> Maybe we can also use a notification displaying some progress, instead of a "Please wait" message :)

Would be nice but given that this is DevView and essentially an only-use-once, maybe not super necessary. I can look into it but it'll have to wait a bit as I'm rather busy atm ^^

PR #448, opened 2023-07-23
chore: Fix typo in comment

GeckoEidechse:
Just fixes a small typo. It's close to an area that #445 touches so I'm gonna merge the other first as I don't know whether the closeness can also cause merge conflict


GeckoEidechse:
Yup called it, this would have caused a merge conflict. Should be resolved now ^^

PR #444, opened 2023-07-23
meta: add Profile management

Jan200101:
meta PR to keep all the not yet PRable commits for profile management.

each and every one is going to be PR'd independenly and the profiles branch will get rebased after a merge.

PR #449, opened 2023-07-24
feat: Primitive dev-only launch params

GeckoEidechse:
- Simply adds a field to DevView to launch with specified launch params. Does not store the launch params in any way.
- This is basically a reduced #419, not to be confused with #301 which is completely separate.

image


GeckoEidechse:
Keeping as draft cause I wanna make separate PR for [restructuring the vuex commit payload](https://github.com/R2NorthstarTools/FlightCore/pull/449/files#diff-32bc04ae938110cf119ad14d3b3c167056516fea1f071eb1d45c1804dbc7ed4dR171-R172)

src-tauri/src/northstar/mod.rs
```diff
@@ -112,8 +113,17 @@ pub fn launch_northstar(
...
+        // We cannot add the params directly because of limitations with cmd.exe
+        // https://stackoverflow.com/questions/9964865/c-system-not-working-when-there-are-spaces-in-two-different-parameters/9965141#9965141
+
+        let launch_parameters = launch_parameters.unwrap_or_default();
+        let ns_params: Vec<&str> = launch_parameters.split_whitespace().collect();
+        dbg!(ns_params.clone());
+        args.extend(ns_params);
         let _output = std::process::Command::new("C:\\Windows\\System32\\cmd.exe")
-            .args(["/C", "start", "", &ns_exe_path])
+            .args(args)
```
GeckoEidechse, line 126: cc @Jan200101 cause this will 100% merge conflict with #444 so we should probably figure out how to best merge it ^^"

I.e. turn profile command into launch arg and then pass it or keep profile arg in `GameInstall` struct and then combine inside the Rust code.
(second one probably makes more sense)

Jan200101: A lot of APIs rely on having profile information available, that would require adapting, which is why I made the profile part of `GameInstall`.

GeckoEidechse: Yeah, profile should stay in `GameInstall` so second option it is then ^^

src-tauri/src/northstar/mod.rs
```diff
@@ -112,10 +113,22 @@ pub fn launch_northstar(
             || matches!(game_install.install_type, InstallType::UNKNOWN))
     {
         let ns_exe_path = format!("{}/NorthstarLauncher.exe", game_install.game_path);
-        let ns_profile_arg = format!("-profile={}", game_install.profile);
 
+        let mut args = vec!["/C", "start", "", &ns_exe_path];
+        let ns_profile_arg = format!("-profile={}", game_install.profile);
+        // We cannot add the params directly because of limitations with cmd.exe
```
catornot, line 119: did you try powershell? how is the same thing for powershell?

Alystrasz:
Closing in favour of #1127.

PR #450, opened 2023-07-24
fix: Move leftover type definition to auto-generated from Rust

GeckoEidechse:
`InstallType` in TypeScript code was still manually created based on the Rust struct. We want this to be auto-generated from Rust code instead to avoid any desync issues if the Rust side gets changed.
(Single source of truth and all that :P)


Alystrasz (approved):
Confirmed working, thanks for the sweep!

Issue #284, opened 2023-04-21
Don't say "Game is running" if it's Northstar dedicated server


GeckoEidechse:
Make sure to check whether `NorthstarLauncher.exe` isn't actually a dedicated server instead.

PR #451, opened 2023-07-25
feat: Add button to forcefully terminate Northstar

Jan200101:
At times Northstar (or rather Titanfall2 in general) locks up completely and needs to be killed.
This adds a button to kill Titanfall2.exe or NorthstarLauncher.exe to the Dev View.

Untested.

### DIFF
--- a/src-tauri/src/util.rs
+++ b/src-tauri/src/util.rs
@@ -74,10 +74,12 @@ pub async fn kill_northstar() -> Result<(), String> {
 
 
     for process in s.processes_by_exact_name("Titanfall2.exe") {
+        log::info!("Killing Process {}", process.pid())
         process.kill();
     }
 
     for process in s.processes_by_exact_name("NorthstarLauncher.exe") {
+        log::info!("Killing Process {}", process.pid())
         process.kill();
     }
 
--- a/src-tauri/src/util.rs
+++ b/src-tauri/src/util.rs
@@ -2,7 +2,7 @@
 
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
-use sysinfo::{SystemExt, ProcessExt};
+use sysinfo::{ProcessExt, SystemExt};
 use zip::ZipArchive;
 
 use crate::constants::{APP_USER_AGENT, MASTER_SERVER_URL, SERVER_BROWSER_ENDPOINT};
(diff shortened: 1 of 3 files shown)


src-tauri/src/util.rs
```diff
@@ -64,6 +64,26 @@ pub async fn get_server_player_count() -> Result<(i32, usize), String> {
...
+pub async fn kill_northstar() -> Result<(), String> {
+    if !check_northstar_running() {
+        return Err("Northstar is not running".to_string());
+    }
+
+    let s = sysinfo::System::new_all();
+
+
+    for process in s.processes_by_exact_name("Titanfall2.exe") {
+        process.kill();
```
GeckoEidechse, line 77: Maybe add a `log::info!("Killing {process}")` before killing it.

src-tauri/src/util.rs
```diff
@@ -64,6 +64,26 @@ pub async fn get_server_player_count() -> Result<(i32, usize), String> {
...
+    }
+
+    for process in s.processes_by_exact_name("NorthstarLauncher.exe") {
+        process.kill();
```
GeckoEidechse, line 81: Ditto

GeckoEidechse (commented):
Looks interesting. Not sure if it can actually kill a stuck Northstar process but we can certainly give it a shot based on recent tickets ^^

src-tauri/src/util.rs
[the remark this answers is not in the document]
Jan200101: Checked the dependency, it only supports sigkill on Windows
https://github.com/GuillaumeGomez/sysinfo/blob/9e1e2684a9d78f3e213dfd8f3382628eab6718ab/src/windows/system.rs#L39-L43

Linux people should know how to kill things anyways :clueless:

src-tauri/src/util.rs
```diff
@@ -64,6 +64,27 @@ pub async fn get_server_player_count() -> Result<(i32, usize), String> {
...
+    }
+
+    let s = sysinfo::System::new_all();
+
+    for process in s.processes_by_exact_name("Titanfall2.exe") {
+        log::info!("Killing Process {}", process.pid());
+        process.kill();
+    }
+
+    for process in s.processes_by_exact_name("NorthstarLauncher.exe") {
```
Alystrasz, line 80: Kinda related to #284: can we make sure this will not kill server instead of game client?

Jan200101: But how would we do that?
NorthstarLauncher only uses the flag to determine if it should load stubs or not and the wsock32_proxy can be ran via Titanfall2.exe

I guess we could get the whole command line and check if it contains `-dedicated`?
If we do that we should put that into a helper function that takes `sysinfo::Process` to re-use for checking if its running and killing.

Since FlightCore does not differentiate between server and client as is right now, I think its fine to keep this and change it up in a follow-up PR

---

What also just came into my head, we would also need to change any launch method to use `-multiple`, since by default Source grabs a named handle to prevent multiple instances from running.

GeckoEidechse: Yeah, I'd skip that tbh. This is on DevView anyway and we primarily target installs with only client running.

GeckoEidechse (approved):
Confirmed working in testing. Launched two instances of Northstar and it killed both.

Same with vanilla Titanfall2.

PR #453, opened 2023-07-29
refactor: Add Profile to GameInstall

Jan200101:
adds a profile attribute to GameInstall.

Should not change any frontend behavior.


GeckoEidechse (commented):
Doing a quick check in vscode, I think there might still be a few spots missing ^^"

image

Checked using the following regex in vscode
```
^(?!.*http).*R2Northstar.*
```

PR #454, opened 2023-07-29
refactor: Replace naive `game_path` argument with `GameInstall`

GeckoEidechse:
Spliced out of #453


GeckoEidechse (commented):
Confirmed working in testing

PR #456, opened 2023-07-30
Extract Northstar into temporary location before installing

Jan200101:
This would allow preprocessing to be done ahead of time, like moving the dll into the profile or renaming the profile.

Nothing is actually done yet, because of some missing bits in Northstar.

### DIFF
--- a/src-tauri/src/northstar/install.rs
+++ b/src-tauri/src/northstar/install.rs
@@ -5,7 +5,7 @@ use std::{cell::RefCell, time::Instant};
 use ts_rs::TS;
 
 use crate::constants::TITANFALL2_STEAM_ID;
-use crate::{util::extract, GameInstall, InstallType};
+use crate::{util::{extract,move_dir_all}, GameInstall, InstallType};
 
 #[cfg(target_os = "windows")]
 use crate::platform_specific::windows;
@@ -40,10 +40,7 @@ async fn do_install(
     let download_directory = format!("{}/download-dir", temp_dir);
     let extract_directory = format!("{}/extract-dir", temp_dir);
 
-    log::info!(
-        "Attempting to create temporary directory {}",
-        temp_dir
-    );
+    log::info!("Attempting to create temporary directory {}", temp_dir);
     std::fs::create_dir_all(download_directory.clone())?;
     std::fs::create_dir_all(extract_directory.clone())?;
 
@@ -100,7 +97,16 @@ async fn do_install(
 
     for entry in std::fs::read_dir(extract_directory).unwrap() {
         let entry = entry.unwrap();
-        let destination = format!("{}/{}", game_path, entry.path().file_name().unwrap().to_str().unwrap().to_string());
+        let destination = format!(
+            "{}/{}",
+            game_path.display(),
+            entry
+                .path()
+                .file_name()
+                .unwrap()
+                .to_str()
+                .unwrap()
+        );
 
         log::info!("Installing {}", entry.path().display());
         if !entry.file_type().unwrap().is_dir() {
--- a/src-tauri/src/northstar/install.rs
+++ b/src-tauri/src/northstar/install.rs
@@ -5,7 +5,10 @@ use std::{cell::RefCell, time::Instant};
 use ts_rs::TS;
 
 use crate::constants::TITANFALL2_STEAM_ID;
-use crate::{util::{extract,move_dir_all}, GameInstall, InstallType};
+use crate::{
+    util::{extract, move_dir_all},
+    GameInstall, InstallType,
+};
 
 #[cfg(target_os = "windows")]
 use crate::platform_specific::windows;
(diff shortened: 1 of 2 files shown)


src-tauri/src/util.rs
```diff
@@ -143,3 +143,38 @@ pub fn check_northstar_running() -> bool {
         || s.processes_by_name("Titanfall2.exe").next().is_some();
     x
 }
+
+pub fn copy_dir_all(
```
GeckoEidechse, line 148: Some doc comments (`///`) would be nice explaining what the function does.

It's kinda obvious/guessable based on file name but it's still good practice to have like a one line comment explaining function behaviour to ensure that assumptions are correct ^^

src-tauri/src/util.rs
```diff
@@ -143,3 +143,38 @@ pub fn check_northstar_running() -> bool {
...
+        if ty.is_dir() {
+            copy_dir_all(entry.path(), dst.as_ref().join(entry.file_name()))?;
+        } else {
+            std::fs::copy(entry.path(), dst.as_ref().join(entry.file_name()))?;
+        }
+    }
+    Ok(())
+}
+
+pub fn move_dir_all(
```
GeckoEidechse, line 165: Ditto

src-tauri/src/util.rs
```diff
@@ -143,3 +143,39 @@ pub fn check_northstar_running() -> bool {
         || s.processes_by_name("Titanfall2.exe").next().is_some();
     x
 }
+
+#[allow(dead_code)]
+pub fn copy_dir_all(
```
Jan200101, line 149: same family of functions as `move_dir_all` generally useful and will become quickly relevant for #444

src-tauri/src/northstar/install.rs
```diff
@@ -36,13 +36,13 @@ async fn do_install(
     game_path: &std::path::Path,
 ) -> Result<()> {
     let filename = format!("northstar-{}.zip", nmod.version);
-    let download_directory = format!("{}/___flightcore-temp-download-dir/", game_path.display());
+    let temp_dir = format!("{}/___flightcore-temp-dir", game_path.display());
```
Jan200101, line 42: since the purpose of the directory is now to do multiple things, the name was changed.
There are many other references to `___flightcore-temp-download-dir` that may need to be adapted, but this could leave leftover files that we can no longer handle when a user did something unexpected in the past.

Should we change the name everywhere, or keep the old name and simply update the structure?

src-tauri/src/util.rs
```diff
@@ -143,3 +143,39 @@ pub fn check_northstar_running() -> bool {
...
+        if ty.is_dir() {
+            copy_dir_all(entry.path(), dst.as_ref().join(entry.file_name()))?;
+        } else {
+            std::fs::copy(entry.path(), dst.as_ref().join(entry.file_name()))?;
+        }
+    }
+    Ok(())
+}
+
+pub fn move_dir_all(
```
Jan200101, line 168: move_dir only moves files, not folders.
So this would result in the folders being left over, but since we delete the whole temp folder it shouldn't matter.

Jan200101:
I love when cargo fmt does one thing and then the CI says another thing. Absolutely lovely

src-tauri/src/util.rs
Jan200101: (this was also mentioned in the comment)

GeckoEidechse:
> I love when cargo fmt does one thing and then the CI says another thing. Absolutely lovely

You got some special formatting settings in [your global config](https://doc.rust-lang.org/cargo/reference/config.html) maybe?

src-tauri/src/northstar/install.rs
GeckoEidechse: If you update the directory name, then make sure to also update the logic called from RepairView that deletes the temp folder to delete both the one with the old name and the new name.

IMO, it would be easiest to just keep the name and have a separate PR for changing it.

src-tauri/src/northstar/install.rs
```diff
@@ -94,11 +94,29 @@ async fn do_install(
...
+            game_install.game_path,
+            entry.path().file_name().unwrap().to_str().unwrap()
+        );
+
+        log::info!("Installing {}", entry.path().display());
+        if !entry.file_type().unwrap().is_dir() {
+            std::fs::rename(entry.path(), destination)?;
+        } else {
+            move_dir_all(entry.path(), destination)?;
+        }
```
GeckoEidechse, line 114: Is there no generic OS move function that works for both files and folders?

Jan200101: None that I found.
fs::rename can do some directory work on Linux, but its semantics make it unreliable.

GeckoEidechse: Aight, I feel like there should be an easier way then having our own code for doing it but if it works, I guess it's good enough for now (until the next refactor :trollface:)

GeckoEidechse (approved):
In my very limited testing I didn't notice any issues so I hope it's fine ^^

src-tauri/src/northstar/install.rs
Jan200101: Directory related APIs tend to be very washy cross platform and standard rust provides nothing outside of creation and deletion.

the `fs_extra` crate provides what we want in a way similar to the stdlib.

PR #457, opened 2023-07-30
refactor: Migrate install related functions to `GameInstall`

Jan200101:
relevant for #444, since we need to know what Northstar Profile to update.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -125,7 +125,6 @@ fn main() {
             verify_install_location,
             get_host_os,
             install_northstar_caller,
-            update_northstar,
             northstar::launch_northstar,
             northstar::launch_northstar_steam,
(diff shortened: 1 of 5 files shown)


src-tauri/src/main.rs
```diff
@@ -329,19 +328,6 @@ async fn install_northstar_caller(
     }
 }
 
-/// Update Northstar install in the given path
-#[tauri::command]
-async fn update_northstar(
```
GeckoEidechse, line 334: TIL that we never called this function in frontend, whoops.

Removal of that function should be separate though IMO.

(I'm running on very little time atm, so the smaller the changes, the larger the chance of merging in reasonable time :P)

GeckoEidechse (approved):
Looking at the changes since last review it should be all good (hopefully)

PR #458, opened 2023-07-30
refactor Generalise temporary directory structure

Jan200101:
temp folder could probably be reused for a lot more things, like extracting Northstar before moving files into the correct place.

### DIFF
--- a/src-tauri/src/repair_and_verify/mod.rs
+++ b/src-tauri/src/repair_and_verify/mod.rs
@@ -42,7 +42,7 @@ pub fn clean_up_download_folder(
 ) -> Result<(), anyhow::Error> {
     // Get download directory
     let download_directory = format!(
-        "{}/___flightcore-temp-download-dir/",
+        "{}/___flightcore-temp/download-dir/",
         game_install.game_path
     );
 
--- a/src-tauri/src/repair_and_verify/mod.rs
+++ b/src-tauri/src/repair_and_verify/mod.rs
@@ -40,9 +40,10 @@ pub fn clean_up_download_folder(
     game_install: &GameInstall,
     force: bool,
 ) -> Result<(), anyhow::Error> {
-    const TEMPORARY_DIRECTORIES: [&str; 2] = [
+    const TEMPORARY_DIRECTORIES: [&str; 3] = [
         "___flightcore-temp-download-dir",
         "___flightcore-temp/download-dir",
+        "___flightcore-temp",
     ];
 
     for directory in TEMPORARY_DIRECTORIES {
--- a/src-tauri/src/repair_and_verify/mod.rs
+++ b/src-tauri/src/repair_and_verify/mod.rs
@@ -40,26 +40,29 @@ pub fn clean_up_download_folder(
     game_install: &GameInstall,
     force: bool,
 ) -> Result<(), anyhow::Error> {
-    // Get download directory
-    let download_directory = format!(
-        "{}/___flightcore-temp/download-dir/",
-        game_install.game_path
-    );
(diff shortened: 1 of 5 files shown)


GeckoEidechse (commented):
Fine by me, just make sure to add the old path to[ `clean_up_download_folder()`](https://github.com/R2NorthstarTools/FlightCore/blob/c73d4862c6a362f443c2db7d93b2cfaf51b42cff/src-tauri/src/repair_and_verify/mod.rs#L39-L64) for cleanup, then it should be ready to merge ^^

https://github.com/R2NorthstarTools/FlightCore/blob/c73d4862c6a362f443c2db7d93b2cfaf51b42cff/src-tauri/src/repair_and_verify/mod.rs#L39-L64

src-tauri/src/repair_and_verify/mod.rs
```diff
@@ -40,26 +40,29 @@ pub fn clean_up_download_folder(
...
     force: bool,
 ) -> Result<(), anyhow::Error> {
-    // Get download directory
-    let download_directory = format!(
-        "{}/___flightcore-temp-download-dir/",
-        game_install.game_path
-    );
+    const TEMPORARY_DIRECTORIES: [&str; 2] = [
+        "___flightcore-temp-download-dir",
+        "___flightcore-temp/download-dir",
```
GeckoEidechse, line 45: from the top of my head, this will result in `___flightcore-temp/` being leftover afterwards

Alystrasz:
@GeckoEidechse can you remind me why we're using a folder inside Northstar's directory, and not creating a temporary folder (*e.g.* in `/tmp` on UNIX) using something like [std::filesystem::temp_directory_path](https://en.cppreference.com/w/cpp/filesystem/temp_directory_path) in cpp?

GeckoEidechse:
> @GeckoEidechse can you remind me why we're using a folder inside Northstar's directory, and not creating a temporary folder (_e.g._ in `/tmp` on UNIX) using something like [std::filesystem::temp_directory_path](https://en.cppreference.com/w/cpp/filesystem/temp_directory_path) in cpp?

Cause when I wrote that code, I didn't know that [`https://doc.rust-lang.org/std/env/fn.temp_dir.html`](https://doc.rust-lang.org/std/env/fn.temp_dir.html) existed and thought I had to find temp dir myself ^^"

Jan200101:
Ignore my misclick

Alystrasz:
> Cause when I wrote that code, I didn't know that https://doc.rust-lang.org/std/env/fn.temp_dir.html existed and thought I had to find temp dir myself ^^"

Let's keep this for another issue then :)

Jan200101:
Having the temp Dir be in the game folder is actually good.

Exctracting Northstar and possible preprocessing it could take some time and moving files on the same fs is cheap, so is the best way to do things.

Alystrasz (approved):
I confirm Northstar installation and launcher PR installation both use the `___flightcore-temp` directory.

Alystrasz:
> Having the temp Dir be in the game folder is actually good.
> 
> Exctracting Northstar and possible preprocessing it could take some time and moving files on the same fs is cheap, fast and atomic, so is the best way to do things.

I'll argue using a "real" temporary directory hides what's happening under the hood to users, which prevents them from deleting the `___flightcore-temp` dir they don't know is useful for, for instance.

But again, it's not the point of this PR :)

Jan200101:
I'll turn it into an issue to discuss possible changing as well as up- and downsides of all variants.

GeckoEidechse (commented):
Didn't test as @Alystrasz already did. Don't see any obvious issues, so hopefully it's fine ^^

PR #464, opened 2023-08-02
feat: Copy release notes to clipboard

GeckoEidechse:
Adds a button to copy release notes to clip board and also auto-copies them when done generating.


GeckoEidechse (commented):
Tested and working fine on my end on Windows. This is a frontend only change so Linux should work just fine as well.

Gonna merge this myself probably when this is done as I'm essentially the sole consumer of the release note generation feature.

PR #465, opened 2023-08-03
Move DLL into Profile for non default Profiles

Jan200101:
This is done so that with profiles we can run different versions of Northstar which would use different versions of the `Northstar.dll`

### DIFF
--- a/src-tauri/src/northstar/install.rs
+++ b/src-tauri/src/northstar/install.rs
@@ -4,7 +4,7 @@ use std::time::Duration;
 use std::{cell::RefCell, time::Instant};
 use ts_rs::TS;
 
-use crate::constants::TITANFALL2_STEAM_ID;
+use crate::constants::{NORTHSTAR_DEFAULT_PROFILE, NORTHSTAR_DLL, TITANFALL2_STEAM_ID};
 use crate::{
     util::{extract, move_dir_all},
     GameInstall, InstallType,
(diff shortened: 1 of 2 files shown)


Jan200101:
Test locally by forcing the branch and hardcoding the profile.

src-tauri/src/northstar/install.rs
```diff
@@ -96,6 +96,22 @@ async fn do_install(
     log::info!("Extracting Northstar...");
     extract(nfile, std::path::Path::new(&extract_directory))?;
 
+    if game_install.profile != "R2Northstar" {
```
GeckoEidechse, line 99: I'd stick the `R2Northstar` string into a constant in `constants.rs` to avoid hardcoding the string in multiple places ^^

src-tauri/src/northstar/install.rs
```diff
@@ -96,6 +96,22 @@ async fn do_install(
...
+        // We are using a non standard Profile, we must:
+        // - move the DLL
+        // - rename the Profile
+
+        // Move DLL into default Profile
+        let old_dll_path = format!("{}/Northstar.dll", extract_directory);
```
GeckoEidechse, line 105: Maybe same with `Northstar.dll`. Or at the very least define it as a variable at the beginning of the codeblock and then just re-use the variable.

src-tauri/src/northstar/install.rs
```diff
@@ -96,6 +96,22 @@ async fn do_install(
...
+        let new_dll_path = format!("{}/R2Northstar/Northstar.dll", extract_directory);
+        std::fs::rename(old_dll_path, new_dll_path)?;
+
+        // rename default Profile
+        let old_profile_path = format!("{}/R2Northstar/", extract_directory);
+        let new_profile_path = format!("{}/{}/", extract_directory, game_install.profile);
+        std::fs::rename(old_profile_path, new_profile_path)?;
```
GeckoEidechse, line 112: Looking just at the diff, I was a bit confused why you are moving the DLL into the R2Northstar profile and then move rename the entire profile folder until I realised that this was still during initial extract process ^^"

Not sure if a comment or sth would help help this make it clearer...

GeckoEidechse (approved):
Confirmed in testing on Windows. With this PR on default profile the `Northstar.dll` is still in the root dir.

When switching to another profile by manually editing the profile entry in `%appdata%\com.github.r2northstartools.flightcore\flight-core-settings.json` before starting FlightCore and then installing Northstar, the `Northstar.dll` is placed in the profile directory instead.

Jan200101:
Same file can be found on Linux on the following path:
`$HOME/.local/share/com.github.r2northstartools/flightcore/flight-core-settings.json`

GeckoEidechse (approved):
Confirmed in testing on Linux. With this PR on default profile the `Northstar.dll` is still in the root dir.

When switching to another profile by manually editing the profile entry in ` $HOME/.local/share/com.github.r2northstartools/flightcore/flight-core-settings.json` before starting FlightCore and then installing Northstar, the `Northstar.dll` is placed in the profile directory instead.

PR #466, opened 2023-08-03
Add extract dir to cleanup

Jan200101:


Jan200101:
Nah, it was added in a PR that was merged later. Nothing you missed.

PR #467, opened 2023-08-03
refactor: Use `GameInstall` object for PR install functions

Jan200101:
Passes the whole `GameInstall` object instead of individual path to functions related to PR installs.

This is done in preparation for #444


GeckoEidechse (approved):
Tested by installing a random mods and launcher PR. The changed code should not affect any current logic. I checked to make sure that all relevant frontend functions that call that code are updated.

PR #469, opened 2023-08-03
fix: Go over all folders during cleanup

GeckoEidechse:
Go over all folders during cleanup instead of early return on error.

If we early return if some folder is non-empty we will skip the remaining folders that might actually be empty.


GeckoEidechse (commented):
Merging myself when CI is done

PR #471, opened 2023-08-03
fix: Give thunderstore mod card images a background color

Jan200101:
Give thunderstore mod card images a background color to prevent alpha issues

**Before:**
image

**After:**
image

Not noticable on images with no alpha, chose lightgrey because of the lack of contrast.

### DIFF
--- a/src-vue/src/components/ThunderstoreModCard.vue
+++ b/src-vue/src/components/ThunderstoreModCard.vue
@@ -312,4 +312,8 @@ export default defineComponent({
     margin-left: 10px;
     height: auto;
 }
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
Can confirm in testing.

GeckoEidechse:
Poking @Alystrasz rq to confirm that this 3 line CSS change is fine, given that you did most of the frontend stuff ^^

Alystrasz (approved):
That is indeed the correct way to set a background color in CSS :)
Thanks for the change.

PR #486, opened 2023-08-05
feat: Strip trailing newline from release notes

GeckoEidechse:
Generating release notes for FlightCore always had two trailing newlines due to the way the code is setup to expect more categories.

Had to remove them every time I made release and it got annoying.

Simply strip the remaining newlines at the end of the release notes to solve that :>


GeckoEidechse (commented):
Dev feature only used by me and tested locally. Merging when CI passes.

PR #487, opened 2023-08-06
feat: Move button for killing NS to RepairView

GeckoEidechse:
Helped someone in tickets so I'm moving it from DevView to RepairView.

image

Translations are only in English and German, gonna leave the rest to Weblate


GeckoEidechse (commented):
Worked in testing locally, merging myself ^^

PR #488, opened 2023-08-07
fix: Ignore missing temp directories on cleanup

Jan200101:
when a directory could not be read it errored out.

This PRs continues instead.


GeckoEidechse (approved):
Obvious fix, skipping testing

PR #493, opened 2023-08-07
fix: try deleting thunderstore package before deleting mod

### LINKED ISSUE #492: Uninstalling mod only deletes the mod directory, not the whole package

Jan200101:
resolves #492 
When deleting a package from the local mod view it only deletes the mods folder.

This PR tries delete the thunderstore mod first and if that fails it tries deleting the mod folder

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -686,7 +686,12 @@ pub fn delete_northstar_mod(game_install: GameInstall, nsmod_name: String) -> Re
         // Installed mod matches specified mod
         if installed_ns_mod.name == nsmod_name {
             // Delete folder
-            return match delete_thunderstore_mod(game_install, installed_ns_mod.thunderstore_mod_string.or(Some("".to_string())).unwrap()) {
+            return match delete_thunderstore_mod(
+                game_install,
+                installed_ns_mod
+                    .thunderstore_mod_string
+                    .unwrap_or("".to_string()),
+            ) {
                 Ok(_) => Ok(()),
                 Err(_) => delete_mod_folder(&installed_ns_mod.directory),
             };
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -679,14 +679,17 @@ pub fn delete_northstar_mod(game_install: GameInstall, nsmod_name: String) -> Re
     }
 
     // Get installed mods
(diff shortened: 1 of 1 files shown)


src-tauri/src/mod_management/mod.rs
```diff
@@ -679,14 +679,22 @@ pub fn delete_northstar_mod(game_install: GameInstall, nsmod_name: String) -> Re
...
     for installed_ns_mod in installed_ns_mods {
         // Installed mod matches specified mod
         if installed_ns_mod.name == nsmod_name {
             // Delete folder
-            return delete_mod_folder(&installed_ns_mod.directory);
+            return match delete_thunderstore_mod(
+                game_install,
+                installed_ns_mod
+                    .thunderstore_mod_string
+                    .unwrap_or("".to_string()),
```
GeckoEidechse, line 693: Uh, I wouldn't pass an empty string (on `None`) to `delete_thunderstore_mod()` cause that will cause it to not match on anything and then cause the legacy version which will panic here:

https://github.com/R2NorthstarTools/FlightCore/blob/264872e0b1d6aa5572fc2f9504d73845505ce973/src-tauri/src/mod_management/legacy.rs#L170-L171

Instead check if `thunderstore_mod_string` is `Some()` and then try to attempt deleting that mod.

src-tauri/src/mod_management/mod.rs
```diff
@@ -679,14 +679,22 @@ pub fn delete_northstar_mod(game_install: GameInstall, nsmod_name: String) -> Re
...
-    let installed_ns_mods = get_installed_mods_and_properties(game_install)?;
+    let installed_ns_mods = get_installed_mods_and_properties(game_install.clone())?;
 
     // Get folder name based on northstarmods
     for installed_ns_mod in installed_ns_mods {
         // Installed mod matches specified mod
         if installed_ns_mod.name == nsmod_name {
             // Delete folder
-            return delete_mod_folder(&installed_ns_mod.directory);
+            return match delete_thunderstore_mod(
```
GeckoEidechse, line 689: This assumes that we always want to delete all Northstar mods from a given Thunderstore package. This might not necessarily reflect what the user actually wants.

Maybe they want [GlitchOverhaul](https://northstar.thunderstore.io/package/S2Mods/GlitchOverhaul/) but only some parts of it for example...

(I swear the whole Northstar mod vs Thunderstore mod thing is such a mess. We should've just switched the mod format when we migrated to Thunderstore x_x)

PR #494, opened 2023-08-08
feat: Add dropdown menu for profiles

Jan200101:
add dropdown menu for profiles

### DIFF
--- a/src-tauri/src/northstar/profile.rs
+++ b/src-tauri/src/northstar/profile.rs
@@ -1,0 +1,73 @@
+use crate::GameInstall;
+
+// These folders are part of Titanfall 2 and
+// should NEVER be used as a Profile
+const SKIP_PATHS: [&str; 8] = [
+    "___flightcore-temp",
+    "__overlay",
+    "bin",
+    "Core",
+    "r2",
+    "vpk",
+    "platform",
+    "Support",
+];
+
+// A profile may have one of these to be detected
+const MAY_CONTAIN: [&str; 10] = [
+    "mods/",
+    "plugins/",
+    "packages/",
+    "logs/",
+    "runtime/",
+    "save_data/",
+    "Northstar.dll",
+    "enabledmods.json",
+    "placeholder.playerdata.pdata",
+    "LEGAL.txt",
+];
+
+#[tauri::command]
+pub fn fetch_profiles(game_install: GameInstall) -> Result<Vec<String>, String> {
+    let mut profiles: Vec<String> = Vec::new();
+
+    for content in MAY_CONTAIN {
+        let pattern = format!("{}/*/{}", game_install.game_path, content);
+        for e in glob::glob(&pattern).expect("Failed to read glob pattern") {
+            let path = e.unwrap();
+            let mut ancestors = path.ancestors();
+
+            ancestors.next();
+
+            let profile_path = std::path::Path::new(ancestors.next().unwrap());
+            let profile_name = profile_path
+                .file_name()
+                .unwrap()
+                .to_os_string()
+                .into_string()
+                .unwrap();
+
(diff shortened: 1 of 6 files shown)


Jan200101:
@GeckoEidechse

Jan200101:
Had to fix a small issue where the first commit did not have a missing import (how this evaded me is unknown, `tauri dev` didn't error on this)

GeckoEidechse (changes requested):
In quick test, switching profile does not updated the detected version in PlayView.

src-tauri/src/northstar/profile.rs
```diff
@@ -0,0 +1,73 @@
...
+    "runtime/",
+    "save_data/",
+    "Northstar.dll",
+    "enabledmods.json",
+    "placeholder.playerdata.pdata",
+    "LEGAL.txt",
+];
+
+#[tauri::command]
+pub fn fetch_profiles(game_install: GameInstall) -> Result<Vec<String>, String> {
```
GeckoEidechse, line 31: I'd add some [short doc](https://doc.rust-lang.org/reference/comments.html#doc-comments) comment to the function.

src-tauri/src/northstar/profile.rs
```diff
@@ -0,0 +1,73 @@
...
+                profiles.push(profile_name);
+            }
+        }
+    }
+
+    Ok(profiles)
+}
+
+#[tauri::command]
+pub fn validate_profile(game_install: GameInstall, profile: String) -> bool {
```
GeckoEidechse, line 60: ditto

src-vue/src/views/SettingsView.vue
```diff
@@ -172,6 +199,35 @@ export default defineComponent({
...
+            let state = store.state;
+
+            await invoke("validate_profile", { gameInstall: state.game_install, profile: value })
+                .then(async (message) => {
+                    if (!message)
+                    {
+                        // Profile is no longer valid, inform the user...
+                        showErrorNotification(
+                            i18n.global.tc('notification.profile.invalid.text'),
+                            i18n.global.tc('notification.profile.invalid.title')
```
GeckoEidechse, line 214: These two translations are missing from `en.json` btw ^^

GeckoEidechse:
UI wise, there should maybe also be some explainer of what profiles are. Maybe a small `?` icon next to `Active Profile` that if hovered/clicked shows a tooltip or reveals small text that gives a short explainer.

Personally I'd still suggest to gate profile selecting behind DevView for now so that we can skip all these small UI quirks to that are kinda needed to make the feature "end-user compatible" and so that we can already test it with contributors.

Then we could even do a 3.0 release where we basically just move it from DevView to RepairView when deemed stable enough ^^

Jan200101:
> In quick test, switching profile does not updated the detected version in PlayView.

I had only picked the first commit, as you requested.
The commits that fix that very cherry picked onto this branch.

Jan200101:
Validated that the verison is now updated when changing the profile by switching to R2Ronin (which FlightCore cannot detect but is still a valid Profile)

Jan200101:
> UI wise, there should maybe also be some explainer of what profiles are.

Future:tm:

> Personally I'd still suggest to gate profile selecting behind DevView

Done

GeckoEidechse (approved):
Confirmed working on Windows and Linux (skipped actually launching the game on Linux cause no working Titanfall2 install)

Requested code changes have been addressed and there isn't anything major I'd change with this PR rn. We could move the `SKIP_PATHS` and `MAY_CONTAIN` to `constant.rs` but I don't really see too much of a need to do so rn ^^

So good to merge from my end :D

Jan200101:
based on #444

PR #498, opened 2023-08-08
fix: Refetch profiles when opening SettingsView

Jan200101:
Refetches profiles every time we open the SettingsView, like the local mod view does.


Jan200101:
based on #444

GeckoEidechse:
> alternatively we could refetch profiles every time we open the SettingsView, like the local mod view does.

Yeah, probably the better way. Unless you need that button specifically I would skip adding it to DevView. When in #444 I mentioned adding buttons to call certain backend functions in spliced out PRs I meant before the frontend logic that actually calls them is implemented ^^"

Jan200101:
alright.
This was inspired by the existing button in the DevView that is kinda pointless:
image


I'll move the profile fetching from store to settingsview mount callback.

GeckoEidechse (approved):
Confirmed working testing on Fedora 38 by creating a new folder in Titanfall2 dir and making it look like a profile folder. Then switching from and to SettingsView and having new profile folder show up.

PR #500, opened 2023-08-08
fix: Only fetch profiles if a game path is available

### LINKED ISSUE #499: Profile related error message pops up when no Titanfall2 install is found
Of course I only think about testing without Titanfall2 folder set after merging #494  (/.-)

image

Jan200101:
fixes #499


GeckoEidechse (approved):
Confirmed working in testing

src-vue/src/plugins/store.ts
```diff
@@ -320,14 +320,16 @@ export const store = createStore<FlightCoreStore>({
...
         async fetchProfiles(state: FlightCoreStore) {
-            await invoke("fetch_profiles", { gameInstall: state.game_install })
-                .then((message) => {
-                    state.available_profiles = message as string[];
-                })
-                .catch((error) => {
-                    console.error(error);
-                    showErrorNotification(error);
-                });
+            if ("game_path" in state.game_install) {
```
GeckoEidechse, line 323: Actually maybe invert the condition and do an early return to reduce nesting ^^"

GeckoEidechse, line 323: And add a comment explaining the check, e.g. something like

```typescript
// Return early if no game path set.
// Missing game path will always cause `fetch_profiles` to face
```

Jan200101: But `if(!(a in b))` is ugly

GeckoEidechse: So is excessive nesting ¯\\\_(ツ)\_/¯

( Also, yet another reason why Rust syntax is superior >:) )

Jan200101:
I consulted this solution and found that it still yields problems.
All of these problems originate from the fact we use GameInstall now all over the place.
I'll look into the best way to solve this without having to duplicate the code 20 dozen times.

src-vue/src/plugins/store.ts
```diff
@@ -61,7 +61,7 @@ export const store = createStore<FlightCoreStore>({
     state(): FlightCoreStore {
         return {
             developer_mode: false,
-            game_install: {} as unknown as GameInstall,
+            game_install: {game_path: "", profile: "", install_type: "UNKNOWN"} as GameInstall,
```
GeckoEidechse, line 64: Can we set `game_path` as `undefined` or something. Setting to `""` feels kinda dangerous. Imagine we modify code to at some point run

```rust
let folder_to_delete = format!("{}/", game_install.game_path);
delete_folder_recursively(folder_to_delete);
```

We really don't wanna [pull a Valve](https://github.com/ValveSoftware/steam-for-linux/issues/3671#issuecomment-70161790) ;)

GeckoEidechse (approved):
Confirmed working in testing on Fedora 38.

Checked behaviour when game install path is not set and when game install path is set.

PR #501, opened 2023-08-08
feat: Add dialog box with a table of all profiles

Jan200101:
this will later be used to manage northstar profiles.

image
image

### DIFF
--- a/src-vue/src/i18n/lang/en.json
+++ b/src-vue/src/i18n/lang/en.json
@@ -11,7 +11,6 @@
         "yes": "Yes",
         "no": "No",
         "error": "Error",
-        "confirm": "Confirm",
         "cancel": "Cancel",
         "informationShort": "Info",
         "downloading": "Downloading",
@@ -116,12 +115,7 @@
             "edit": "Edit Profiles",
 
             "dialog": {
-                "title": "Profiles",
-                "delete_confirm": "Are you sure to delete this profile?",
-                "delete": "Delete",
-                "clone": "Clone",
-                "new_profile_name": "Enter the new Profile name",
-                "create_empty": "New Profile"
+                "title": "Profiles"
             }
         },
 
--- a/src-vue/src/i18n/lang/en.json
+++ b/src-vue/src/i18n/lang/en.json
@@ -11,6 +11,7 @@
         "yes": "Yes",
         "no": "No",
         "error": "Error",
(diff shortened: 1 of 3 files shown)


src-vue/src/i18n/lang/en.json
```diff
@@ -11,6 +11,7 @@
         "yes": "Yes",
         "no": "No",
         "error": "Error",
+        "confirm": "Confirm",
```
GeckoEidechse, line 14: Technically that translation isn't in the code of this PR... ^^"

src-vue/src/i18n/lang/en.json
```diff
@@ -111,7 +112,17 @@
...
+            "active": "Active Profile",
+            "edit": "Edit Profiles",
+
+            "dialog": {
+                "title": "Profiles",
+                "delete_confirm": "Are you sure to delete this profile?",
+                "delete": "Delete",
+                "clone": "Clone",
+                "new_profile_name": "Enter the new Profile name",
+                "create_empty": "New Profile"
```
GeckoEidechse, line 124: Ditto

GeckoEidechse (commented):
Might be a bit too pedantic here... ^^

GeckoEidechse (commented):
So a thing I noticed with this where honestly I'm not sure how to fix is that pop-up affects the whole "page" which in our case includes the menu bar which means that for as long as the pop-up is open, the menu bar elements cannot be interacted with.

image

I have no idea how to fix that so uh @Alystrasz any ideas? ^^"

GeckoEidechse:
> How do you expect to make the menu bar separate from the dialog?

I have no idea, hence I tagged Alystrasz. I'm sure there must a way to separate main menu bar from window but I don't know how. Alternatively we could bring back native window controls, @Alystrasz :>

Alystrasz:
I guess clicking outside the dialog closes it, so fine by me.
The point of the dialog is to attract user attention to its content, so it's ok if they can't use the menu bar.

GeckoEidechse:
> I guess clicking outside the dialog closes it, so fine by me.

Huh, I could've sworn it didn't in initial testing. Just retried again on Linux and it closes again when clicking outside so yeah, should be fine then ^^

GeckoEidechse (approved):
Works in testing, issue about menu bar also being affected by pop-up was resolved in discussion.

PR #533, opened 2023-08-28
Allow ProtonGE to be used as a compatibility tool

Ombrezz:
Both NorthstarProton and ProtonGE work with Northstar, so FlightCore should allow both of them to be used when launching.


GeckoEidechse:
For CI, you can ignore any of "useless vec" errors. Those are unrelated to the PR and caused by #538 which I need to fix myself ^^"

Jan200101:
looks good to me.
Its a bit hard to check for the GE version, because 8-11 has it, 8-12 doesn't and 8-13 has it again.

Ombrezz:
Just updated to use a constant list of valid Proton versions, does this work better?

GeckoEidechse:
Might be better yeah. Can you update branch to newest `main`? I think I fixed all the CI issues on `main` now ^^

GeckoEidechse:
Heads-up that some files fail format-check

Ombrezz:
Anymore work that needs to be done before this can get merged?

GeckoEidechse:
> Anymore work that needs to be done before this can get merged?

Nothing I can see on a quick skim. I'm just dealing with a massive backlog rn so I didn't manage to get to this PR yet...

src-tauri/src/constants.rs
```diff
@@ -58,3 +58,7 @@ pub const NORTHSTAR_DLL: &str = "Northstar.dll";
 
 // Profile that Northstar defaults to and ships with
 pub const NORTHSTAR_DEFAULT_PROFILE: &str = "R2Northstar";
+
+// List of valid compatibility tools that Northstar can be launched with
```
GeckoEidechse, line 62: I switched to doc comments in 7f26fc545503c523a6ce2bf2fc198190ec5720a3 so this should also be updated accordingly ^^

```suggestion
/// List of valid compatibility tools that Northstar can be launched with
```

src-tauri/src/northstar/mod.rs
```diff
@@ -145,15 +145,11 @@ pub fn launch_northstar_steam(
...
                 match steamdir.compat_tool(&titanfall2_steamid) {
                     Some(compat) => {
-                        if !compat
-                            .name
-                            .clone()
-                            .unwrap()
-                            .to_ascii_lowercase()
-                            .contains("northstarproton")
+                        if !VALID_NORTHSTAR_PROTON_BUILDS
+                            .contains(&compat.clone().name.unwrap().as_str())
```
GeckoEidechse, line 149: @Jan200101 is changing the check to also include casing an issue?

Or to say it differently, was there a reason to ignore case sensitivity in previous version? ^^

Ombrezz: You'd have to ask @Jan200101, but I think it may have just been there for convenience. During development I had it log the strings of the various proton builds (including northstarproton) and they all matched the name of the folders that they were in, whether that was a coincidence or not, I don't know. But if you follow the standard installation instructions for these versions, it should match up fine without the casing check. If you want to be extra safe, it can be added back in.

Jan200101: it don't matter, I just kinda did it to give it a little room for error

src-tauri/src/constants.rs
```diff
@@ -58,3 +58,8 @@ pub const NORTHSTAR_DLL: &str = "Northstar.dll";
...
+pub const VALID_NORTHSTAR_PROTON_BUILDS: [&str; 3] = [
+    "NorthstarProton-8.1-1", "GE-Proton8-13", "GE-Proton8-11"
+];
```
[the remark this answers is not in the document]
Ombrezz: This is for an older commit? Do you want a newline on the latest one?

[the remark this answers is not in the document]
GeckoEidechse: whoops no. I still had that review queued from when I took a look at the PR way earlier, sorry.

src-tauri/src/northstar/mod.rs
Ombrezz: Sounds like the check is good as it is, marking as resolved.

GeckoEidechse (approved):
I don't really have a way to test this atm, so I'm just gonna assume this works and will let you double check post-release @Ombrezz ^^

PR #547, opened 2023-09-08
fix: Remove hashes around raw string literal

GeckoEidechse:
Remove hashes around raw string literal clippy said they are "unnecessary"

Should resolve current CI failure


GeckoEidechse (commented):
Clippy no longer fails and this change has no effect on resulting binary so I'm merging without testing if CI passes fully.

PR #548, opened 2023-09-08
i18n: Enable Spanish translations

GeckoEidechse:
Adds _Spanish_ to the language dropdown menu in settings.

image


GeckoEidechse (commented):
Tested locally on Linux before pushing and worked fine.

PR #549, opened 2023-09-08
chore: Delete empty Vietnamese translation file

GeckoEidechse:
Someone requested Vietnamese as a language for FlightCore on Weblate but then never did any translations.
As such'm removing the empty file again for cleanup


GeckoEidechse (commented):
Deletes unused file, merging without testing.

PR #561, opened 2023-09-12
fix: Turn comments into doc comments

GeckoEidechse:
So for some reason when I started out writing this file I used standard comments to annotate the constants.

Using doc comments makes more sense here as they get taken into account when generating documentations.


GeckoEidechse (commented):
This only changes comments, no testing needed.

PR #562, opened 2023-09-13
fix: Ignore NS dedicated server for game running check

### LINKED ISSUE #284: Don't say "Game is running" if it's Northstar dedicated server

Ombrezz:
`check_northstar_running()` only returns true if Northstar wasn't launched with the `-dedicated` flag, untested on Windows.

Closes #284


GeckoEidechse (commented):
Many thanks for the PR <3

I enabled CI and it failed [clippy check](https://doc.rust-lang.org/stable/clippy/usage.html). The fix should be [explained in CI logs](https://github.com/R2NorthstarTools/FlightCore/actions/runs/6166903897/job/16749200116#step:7:815).

If you have any questions, LMK ^^

src-tauri/src/util.rs
```diff
@@ -136,12 +136,14 @@ pub fn check_ea_app_or_origin_running() -> bool {
 /// Checks if Northstar process is running
```
GeckoEidechse, line 136: Would also update the comment

```suggestion
/// Checks if Northstar process is running (ignores dedicated server)
```

GeckoEidechse (changes requested):
So in quick testing on Windows, this cause non-dedicated-Northstar to be detected as running.

Both when starting first a dedi server and then client, and when only starting a client.

Not quite sure what's causing that cause looking at the code it should be fine?

Ombrezz:
The `sysinfo` crate [suggests](https://docs.rs/sysinfo/latest/sysinfo/index.html#usage) `refresh_all()` on a new `System` struct to get all the information up to date, maybe that will fix it?

GeckoEidechse:
Nope, looks like it still has the same issue... :/

image

Did it work for you locally?

Ombrezz:
Works fine for me on Linux, could you run a quick test to see what `sysinfo` returns for the `name()`, `exe()`, and `cmd()` functions on the Northstar and/or TF2 processes on Windows?

src-tauri/src/util.rs
```diff
@@ -175,15 +175,18 @@ pub fn check_ea_app_or_origin_running() -> bool {
...
-    let x = s
-        .processes_by_name("NorthstarLauncher.exe")
-        .next()
-        .is_some()
-        || s.processes_by_name("Titanfall2.exe").next().is_some();
-    x
+    let mut s = sysinfo::System::new_all();
+    s.refresh_all();
+    for process in s.processes().values() {
+        if (process.name().ends_with("Titanfall2.exe") || process.name().ends_with("Northstar.exe"))
```
Alystrasz, line 183: Related to https://github.com/R2NorthstarTools/FlightCore/pull/562#issuecomment-1723204223 I think, process name was changed in this PR while it shouldn't have been modified:
```suggestion
        if (process.name().ends_with("Titanfall2.exe") || process.name().ends_with("NorthstarLauncher.exe"))
```

Alystrasz (changes requested):

Alystrasz:
Closing as stale.

PR #600, opened 2023-10-09
build: Add npm packge in root dir to bot config

GeckoEidechse:
The repo root dir also has a `package.json` which contains the necessary build tools. We wanna keep those up-to-date as well.


GeckoEidechse (commented):
Merging without testing as this just changes the dependabot config.

PR #604, opened 2023-10-09
chore: Bump tauri version

GeckoEidechse:
Combines:
- #596 
- #598 
- #601 
- #602 

into a single PR as all the libraries are more or less dependent on each other.


GeckoEidechse (commented):
Works in both building and testing on Windows 10.

PR #607, opened 2023-10-10
feat: Remove compatibility tool check

Jan200101:
All major proton versions can run Northstar out of the box or with minor configurations:
- Proton 8 needs a DLLOVERRIDE
- GE-Proton works like a charm with a special Northstar case
- NorthstarProton always worked


GeckoEidechse (approved):
So I can confirm that with this change it does launch Northstar even under Proton experimental.

After intro screen, Northstar did crash on Fedora 38 but that's probably some compat issue, so unrelated to FlightCore.

GeckoEidechse:
We should however look into adding functionality to FlightCore for applying the minor required configuration changes (semi-)automatically.

PR #610, opened 2023-10-11
fix: Update popup style

### LINKED ISSUE #608: Fix word break in profile delete dialog
_Originally posted by @Alystrasz in https://github.com/R2NorthstarTools/FlightCore/pull/603#discussion_r1350702604_

> You should update the style of this dialog so that there aren't line breaks in words:
> 
> image

Alystrasz:
Makes popup dialog windows slightly larger and prevents words from breaking.

Closes #608.

##### Before

image

##### After

image

### DIFF
--- a/src-vue/src/style.css
+++ b/src-vue/src/style.css
@@ -46,3 +46,11 @@ body {
     color: white;
     margin: 30px 15px;
 }
+
+.el-popper {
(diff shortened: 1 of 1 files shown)


GeckoEidechse (approved):
Code changes look fine and change looks good in testing ^^

PR #614, opened 2023-10-12
refactor: Use fully qualified path for vars and funcs

GeckoEidechse:
Use fully qualified path for variable instead of importing it directly


GeckoEidechse:
Code refactoring, if CI passes this is good to merge

PR #615, opened 2023-10-12
feat: Notification menu

### LINKED ISSUE #400: Add a notifications button
When installing mod PRs, etc i quite often alt+tab to read the PR descriptions and miss the notification that the PR has been successfully installed. While im sure your code is robust and wouldn't fail, it would be a nice QOL feature to be able to check previous notifications.

Something, something like this:
image

Alystrasz:
This PR introduces a notification menu, which hosts notifications fired while the app was not focused.
Notifications can be closed by the user.

Closes #400.

##### Screenshots

image

image

### DIFF
--- a/src-vue/src/App.vue
+++ b/src-vue/src/App.vue
@@ -34,9 +34,6 @@ export default {
     this.$root!.$i18n.locale = lang;
   },
   methods: {
-    async toggleMaximize() {
-      await appWindow.toggleMaximize();
-    },
     minimize() {
       appWindow.minimize()
     },
@@ -77,7 +74,6 @@ export default {
       <!-- Window controls -->
       <div id="fc_window__controls">
         <el-button color="white" icon="SemiSelect" @click="minimize" circle />
-        <el-button color="white" icon="FullScreen" @click="toggleMaximize" circle />
         <el-button color="white" icon="CloseBold" @click="close" circle />
       </div>
     </nav>
--- a/src-vue/src/App.vue
+++ b/src-vue/src/App.vue
@@ -8,9 +8,11 @@ import { appWindow } from '@tauri-apps/api/window';
 import { store } from './plugins/store';
 import { Store } from 'tauri-plugin-store-api';
 import { invoke } from "@tauri-apps/api";
+import NotificationButton from "./components/NotificationButton.vue";
 
 export default {
   components: {
+      NotificationButton,
       ChangelogView,
       DeveloperView,
       PlayView,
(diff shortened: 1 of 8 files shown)


GeckoEidechse:
Ooooh, me likey 👀

src-vue/src/App.vue
```diff
@@ -34,9 +36,6 @@ export default {
     this.$root!.$i18n.locale = lang;
   },
   methods: {
-    async toggleMaximize() {
-      await appWindow.toggleMaximize();
-    },
```
GeckoEidechse, line 39: Is the reason for removing maximise to make space for the notification icon? ^^

Alystrasz: Yep, and also because I don't like the icon + window can still be maximized by double-clicking the menu bar ;)

GeckoEidechse (commented):
So far worked fine in initial testing :D

GeckoEidechse:
Tagging @F1F7Y for review as well as they requested the feature

src-vue/src/App.vue
GeckoEidechse: Aight, in that case I'll just make a separate PR with that change rq to get the diff down a bit more ^^

GeckoEidechse: #617

GeckoEidechse (approved):
Worked fine in testing for me. Pressed a bunch of buttons that trigger notifications and alt-tabbed before they displayed to get said notifications sent to notification icon.

src-vue/src/utils/ui.ts
```diff
@@ -1,15 +1,21 @@
...
+ * If the app is not focused when this is invoked, a notification is added to the notifications menu.
  **/
 function showNotification(
     title: string,
     message: string = '',
     type: 'success' | 'warning' | 'error' | 'info' = 'success',
     duration: number = 4500
 ): NotificationHandle {
+    if (!document.hasFocus()) {
+        store.commit('addNotification', {title, text: message, type});
```
GeckoEidechse, line 16: I think there's probably also something in the Tauri API to make the taskbar icon flash, although I'm fine with moving that feature to a separate PR \^\^

GeckoEidechse (approved):
Code looks fine as far as I can tell. Didn't see any obvious issues or the like \^\^

src-vue/src/utils/ui.ts
GeckoEidechse: Moved to #634

PR #619, opened 2023-10-13
refactor: Move NS installation logic to module

GeckoEidechse:
Move Northstar installation logic to module in order to clean up `main.rs`


GeckoEidechse (commented):
Refactor PR just moving logic so no testing needed as long as CI passes.

PR #622, opened 2023-10-13
refactor: Move repair/verify related funcs to mod

GeckoEidechse:
Moves functions related to game install path verification checks to dedicated module in an effort to clean up `main.rs`


GeckoEidechse (commented):
Refactor PR that just moves logic around in source files. Merging when CI passes.

PR #629, opened 2023-10-13
refactor: Move `convert_release_candidate_number`

GeckoEidechse:
Move `convert_release_candidate_number` to `util` module to clean up `main.rs`


GeckoEidechse (commented):
Refactor PR that just moves logic around. If CI passes this is good to merge

PR #644, opened 2023-10-24
fix: UI bugs

Alystrasz:
Fixes two interface issues caused by leaking CSS styles:

#### Before

Fucked notification

#### After

Fixed notification

#### Before

Main fucked

#### After

Main OK

#### Before

profiles-cut

#### After

Profiles OK

#### Before

languages-fucked

#### After

Languages OK


GeckoEidechse (approved):
Confirmed in testing.

PR #651, opened 2023-10-28
fix: Address broken rc number conversion

GeckoEidechse:
Address broken release-candidate number conversion for double digit patch releases

`v1.19.10-rc1` would get converted to `v1.19.11` instead of `v1.19.1001`.

Tested with #652


GeckoEidechse (commented):
Tested myself as working on Fedora 38 with `v1.19.10-rc1` of Northstar

PR #652, opened 2023-10-28
test: Add unit tests for rc number conversion

GeckoEidechse:
Adds unit test for converting release candidate numbers.

~~Does not actually run the unit tests in CI. That's gonna be a separate PR.~~ Turns out I was already running unit tests for TypeScript file types generation lmao


GeckoEidechse (commented):
Accidentally already tested what happens in CI when tests fail and tests are not compiled into release app so there's no point in testing release build locally. ^^

PR #669, opened 2023-11-09
fix: Proper release-candidate number conversion

GeckoEidechse:
Addresses:
https://github.com/R2NorthstarTools/FlightCore/blob/2aac1889a7fb8689f8b45e4de5eb8c73ae61eba3/src-tauri/src/util.rs#L232-L234


Alystrasz (changes requested):
I wouldn't mind additional tests, using more than two digits for rc identifier for instance

GeckoEidechse:
> I wouldn't mind additional tests, using more than two digits for rc identifier for instance

More than two digits on release candidate number will never happen cause it already breaks stuff on Northstar upstream.
I.e.: `v1.2.3-rc456` would become `v1.2.3456` which is indiscernible from `v1.2.34-rc56` which also gets translated to `v1.2.3456` ^^

I can add a test for the latter case though.

GeckoEidechse:
Done with f7777d38b5560f17e2ec86e67d7598285768ceb2

PR #620, opened 2023-10-13
chore: Properly populate package manifest info

GeckoEidechse:
Closes #580

PR #672, opened 2023-11-10
build: Fix pdb not uploading to sentry

GeckoEidechse:
The filename was changed in #620 but the CI config was never updated accordingly.
As such it still tried to upload the old filename and would silent fail.


GeckoEidechse:
Release CI config change, we don't need to wait for build CI to finish.

Issue #752, opened 2024-01-17
Overhaul dev launching option


GeckoEidechse:
In DevView, have like a bunch of checkboxes like
- skip update check
- check EA app running
- launch via Steam

etc

and a single launch button instead of a button for each each combination of option like we kinda have rn.

PR #753, opened 2024-01-17
refactor: Make launching via Steam an arg passed to backend

GeckoEidechse:
Make launching via Steam an arg passed to backend instead of a separate dedicated function.

Part of #752


GeckoEidechse (commented):
Tested and confirmed working on my end on Windows

PR #760, opened 2024-01-19
refactor: Pass launch options via object

GeckoEidechse:
Currently object only has a single member for bypassing all checks. In the future this would be split up into multiple members for different checks to bypass as well as other options.

Then I can get rid of the `launchGameViaSteam` function and just make it some bool passed to backend. In fact, the backend has already been adjusted for this ^^

The goal is to have something like this, where different options for how to launch Northstar could be set.

image

Maybe even extended with some degree of launch args, primarily focused on dev stuff, so more towards #449 and less #301 (although the same object could be extended to pass launch args ^^

Requires:
- #798

Part of:
- #752


GeckoEidechse:
Requesting review from @Alystrasz cause it's been a while since I last wrote TypeScript and I feel like I'm using rather primitive patterns that probably have a smarter way of doing things.

src-vue/src/plugins/store.ts
```diff
@@ -172,7 +173,13 @@ export const store = createStore<FlightCoreStore>({
                 }
             }
         },
-        async launchGame(state: any, no_checks = false) {
+        async launchGame(state: any, launch_options: LaunchOptions | null = null) {
+            let no_checks = false;
+
+            if (launch_options != null) {
+                no_checks = launch_options.no_checks;
```
Alystrasz, line 180: Instead of manually extracting `LaunchOptions` members one by one, couldn't we pass options directly to backend using a Rust struct and its associated Typescript binding?

Alystrasz (changes requested):

src-vue/src/plugins/store.ts
GeckoEidechse: Yeah, that was the longterm goal but I realise now, that I might as well do it already from the get-go ^^

GeckoEidechse: -> #798 for passing the object directly to backend as a whole ^^

GeckoEidechse:
This should be ready for re-review now ^^

GeckoEidechse (commented):
In quick testing on Linux I can confirm that the changed front-end logic for launching via Steam and with bypassing checks still works.

Cannot test the standard play button atm though...

Alystrasz (approved):
Comments addressed, confirmed working in testing.

PR #761, opened 2024-01-20
feat: Add option to clone existing profile

GeckoEidechse:
Basically is just #444 repurposed and as such replaces #444

image


src-vue/src/views/DeveloperView.vue
```diff
@@ -74,6 +74,10 @@
                 Get installed mods
             </el-button>
 
+            <el-button type="primary" @click="getAvailableProfiles">
```
Alystrasz, line 77: Is this of any use?

src-tauri/src/main.rs
```diff
@@ -141,6 +141,8 @@ fn main() {
             northstar::profile::delete_profile,
             northstar::profile::fetch_profiles,
             northstar::profile::validate_profile,
+            northstar::profile::delete_profile,
```
Alystrasz, line 144: Why did you put that in the tauri handler, since profiles can already be deleted?

src-vue/src/views/DeveloperView.vue
GeckoEidechse: Whoops, no. Good catch. Was some leftover code. Removed with 2000de537d17e9b547cdd7d8466d7b863e1bfbb9

src-tauri/src/main.rs
GeckoEidechse: Was a leftover from a bad merge, removed with 275acb7e04a34f7d874cd13506537a780b36188e o7

Alystrasz (approved):
Confirmed working in testing.
Profiles can be cloned, and cannot use naming of an already-existing profile.

PR #766, opened 2024-01-24
chore: Bump checkout action to v4

GeckoEidechse:
v3 uses Node 16 which is slated for deprecation by GitHub Actions

image


GeckoEidechse:
Not much to test here. If CI passes this is good to merge ^^

PR #787, opened 2024-02-06
feat: Retry package index fetch on failure

GeckoEidechse:
Retry package index fetch on failure with wait time in-between and max number of retries.


GeckoEidechse:
Keeping as draft for now as I'm not sure if this is the right approach or whether it'll actual improve some of the errors we are seeing.

GeckoEidechse:
Did not work to resolve the connection issues experienced by players in tickets on the Northstar Discord.

PR #790, opened 2024-02-10
feat: Add FlightCore SVG

cyrv6737:
Traced by hand because bitmap tracing sucks. Blame any symmetry issues on Respawn lol


GeckoEidechse:
Clippy failing on something unrelated to the change is a mood xD

tbf, it's cause the Rust toolchain in the CI continuously updates and as such also pulls in newer and more capable versions of clippy that can spot previous undetected bad patterns. Still funny though. Will fix in a sec ^^

GeckoEidechse:
Thanks for creating the SVG <3

PR #792, opened 2024-02-11
feat: Release announcements generator

GeckoEidechse:
Adds dev tool to semi-auto-generate the release announcements I post in `#releases` on the Northstar Discord server.

The idea being that this should help me in reducing some of manual labour required to creating releases.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -99,6 +99,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arc-swap"
+version = "1.6.0"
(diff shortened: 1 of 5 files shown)


GeckoEidechse:
Marked as draft while I polish it a bit more ^^

GeckoEidechse:
There's a bunch of `unwraps` in the code still but given that this is developer code I don't really care about potential crashes tbh. It's only getting called by me and the crashes only happen when it fails to connect to GitHub.

GeckoEidechse:
This should be ready now. Given that it is a development tool, I will merge it myself once CI has passed.

PR #796, opened 2024-02-12
refactor: Remove unused function arg

GeckoEidechse:
`bypass_checks` was never used for `launch_northstar_steam`


GeckoEidechse:
No testing needed as the function is only called from backend so Rust compiler should throw an error if there would be any code that wasn't updated accordingly.

PR #798, opened 2024-02-12
refactor: Pass object to backend instead of individual args

GeckoEidechse:
Instead of passing individual args to `launch_northstar` in the backend, pass a single object. The idea here being that said object can later easily be extended with more fields such as launch args for Northstar


GeckoEidechse:
The object is still freshly created in `store.ts` instead of being passed to the relevant frontend function. That part will be refactored in #760

GeckoEidechse:
Fuck it, we test in prod :>

PR #803, opened 2024-02-13
fix: Remove most leftover dbg prints

GeckoEidechse:
Removes all but one leftover `dbg!()` prints. The remaining one is in a function that is yet to be implemented ^^


GeckoEidechse:
Nothing to review here, merging

Issue #806, opened 2024-02-13
Replace manual GitHub interactions with library


GeckoEidechse:
https://crates.io/crates/octocrab

PR #807, opened 2024-02-13
refactor: Use dedicated library for getting PR list

GeckoEidechse:
Use the [`octocrab`](https://crates.io/crates/octocrab) library for interacting with GitHub for getting the list of pull requests

Part of:
- #806


GeckoEidechse (commented):
Tested locally and was still able to grab PR list.

PR #808, opened 2024-02-13
refactor: Use `u64` for PR number

GeckoEidechse:
Use `u64` for PR number instead of `i64` as PR numbers can never be negative.

Discovered while working on #807


ASpoonPlaysGames:
a `u64`? how many PRs do you expect to be made? a u16 would honestly be enough

GeckoEidechse:
> a `u64`? how many PRs do you expect to be made? a u16 would honestly be enough

You underestimate how many PRs I can get stuck in PR hell >:)

&nbsp;

Jokes aside, I'm just matching [what the library I'm about to switch to uses](https://docs.rs/octocrab/0.34.1/octocrab/models/pulls/struct.PullRequest.html#structfield.number), haha ^^"

Issue #340, opened 2023-05-09
Add screenshots to weblate


GeckoEidechse:
To help translators we should add screenshots of the English interface of FlightCore to [weblate](https://translate.harmony.tf/).

PR #819, opened 2024-02-21
i18n: Translations update from Weblate

harmony-weblate:
Translations update from [Weblate](https://translate.harmony.tf) for [Northstar/FlightCore](https://translate.harmony.tf/projects/northstar/flightcore/).

Current translation status:

Weblate translation status

### DIFF
--- a/src-vue/src/i18n/lang/pl.json
+++ b/src-vue/src/i18n/lang/pl.json
@@ -126,7 +126,7 @@
                 "delete": "Usuń",
                 "clone": "Klonuj",
                 "new_profile_name": "Wprowadź nową nazwę profilu",
-                "create_empty": "Nowy Profil",
+                "create_empty": "Nowy profil",
                 "delete_confirm": "Czy na pewno chcesz usunąć ten profil?"
             },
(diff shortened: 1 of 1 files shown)


p0358:
Actually I only corrected two strings on translations made by someone else xd

GeckoEidechse:
Oh, it's two commits, lol. I assume the translations are fine?

p0358:
Yeah, except I have a serious doubt about `"date_prefix": "na"`

You should really make use of that screenshots function, because it's impossible to know what's the right translation for many of these things without seeing where they're actually used (and that's obvious only to the developers of the actual software, but translators can have no clue)

So I'm almost sure that one particular translation will be wrong, but correct one depends on where and how it's used...

GeckoEidechse:
Aighty, I'll still merge then ig cause reverting translations is a bit of a hassle x_x

I'd love to add screenshots to weblate but I got other stuff to work on that's a bit more pressing. Made an issue in via #340 so if anyone wants to help out, get in touch and we'll set up perms or sth ig.

PR #871, opened 2024-03-29
feat: Add disable all mods button to local mod view

GeckoEidechse:
Initial PoC to add a button to local mod view to directly disable all mods except core ones.

This button already exists in RepairView but given how cumbersome that is to access, having a button additionally directly in the local mods view would be nice.

For context I observed a lot of cases recently on the Northstar Discord where players would have their game break by conflicting mods and staff would tell them to try to disable individual mods one-by-one. I would argue that instead having a quick way to disable all mods and then slowly re-enabling them would be a way faster method of reducing the player's issue.

This PR adds a button to achieve just that ^^

image

### DIFF
--- a/src-vue/src/views/mods/LocalModsView.vue
+++ b/src-vue/src/views/mods/LocalModsView.vue
@@ -5,7 +5,7 @@
     </div>
 
     <el-scrollbar v-else>
-        <el-button type="primary" @click="disableAllModsButCore">
+        <el-button class="disableModsBtn" type="primary" @click="disableAllModsButCore">
             {{ $t('settings.repair.window.disable_all_but_core') }}
         </el-button>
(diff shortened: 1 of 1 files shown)


GeckoEidechse:
Obviously the styling/padding looks kinda meh, requesting assistance @Alystrasz <3

GeckoEidechse:
Looks better now, thanks @Alystrasz <3

[Screencast from 2024-04-03 17-43-31.webm](https://github.com/R2NorthstarTools/FlightCore/assets/40122905/9c06a1ea-3b4e-4848-b8e2-ed409ecf49f0)

GeckoEidechse:
With the changes done by @Alystrasz this should be good to merge now IMO :D

PR #893, opened 2024-04-14
chore: Remove Linux checks

Jan200101:
With NorthstarProton changing to be GEProton based the glibc requirement is no longer needed since it targets the steam runtime


GeckoEidechse (approved):
Linux related change. Skipping testing. This should not affect Windows.

PR #894, opened 2024-04-14
chore: Update tauri-plugin-store

Jan200101:
uses the dependency declaration recommended by upstream and lets us get rid of ugly build errors

image

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -4721,8 +4721,8 @@ dependencies = [
 
 [[package]]
 name = "tauri-plugin-store"
-version = "0.1.0"
-source = "git+https://github.com/tauri-apps/plugins-workspace?rev=5a6abd3203dc94c38f96d0c4bf7ecbef399f8c25#5a6abd3203dc94c38f96d0c4bf7ecbef399f8c25"
+version = "0.0.0"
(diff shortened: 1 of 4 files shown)


GeckoEidechse:
Didn't know they finally had a release, thanks :D

Do we also need to update the `package.json` for it as well?

https://github.com/R2NorthstarTools/FlightCore/blob/7ab55d005aa7661d1eed6cb0a77ee6b269bfdf91/src-vue/package.json#L15

Jan200101:
> Didn't know they finally had a release, thanks :D

They don't
and they don't appear to be versioning the v1 version at all
but I don't want to look at that stupid error anymore and the v1 release appears to be left as is with only fixes no new features or changes.

> Do we also need to update the package.json for it as well?

fuck
give me a moment

Jan200101:
okay, bumped the js package and made it follow the v1 release
On that note, the JS dependency was hardcoded to https://github.com/tauri-apps/tauri-plugin-store/commit/9bd993aa67766596638bbfd91e79a1bf8f632014 a 2022 commit
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
this also bumps the nodejs requirement to 14.6 but completely drops tslib

GeckoEidechse (approved):
Confirmed working in testing on Windows by changing language and restarting the application to see if the change persists.

PR #902, opened 2024-04-21
fix: Properly handle steamlocate app location tristate

Jan200101:
`find_app` returns a complex a `Result<Option<t>>` which gives us the tristate of 

1. checked and its there
2. checked and its not there
3. something fucked up

we only caught 1 and 3


GeckoEidechse (approved):
Should address https://northstar-kv.sentry.io/issues/5234059930/events/c1d10a43af454c0b97ab005045a3fb20/

PR #931, opened 2024-05-29
i18n: Translations update from Weblate

harmony-weblate:
Translations update from [Weblate](https://translate.harmony.tf) for [Northstar/FlightCore](https://translate.harmony.tf/projects/northstar/flightcore/).

Current translation status:

Weblate translation status


GeckoEidechse:
Given that these translations seem to have been done using an auto-translation tool like Google Translate, I'm rejecting them and will be resetting them on weblate.

PR #949, opened 2024-06-18
fix: if multiple artifacts are available check for MSVC build

Jan200101:
needed for https://github.com/R2Northstar/NorthstarLauncher/pull/721

Is backwards compatible


GeckoEidechse (approved):
Confirmed in quick testing by installing https://github.com/R2Northstar/NorthstarLauncher/pull/721 and https://github.com/R2Northstar/NorthstarLauncher/pull/722 and FlightCore placing the `.exe` and `.dll` correctly in both cases ^^

PR #950, opened 2024-06-18
fix: Use border when focusing menu items

Jan200101:
Element Plus implements the menu using lists which do not render outlines, which all other objects use to implement the focus outline.

Also the dev menu entry is not tab indexed for some reason?
Element Plus just kinda sucks.

### DIFF
--- a/src-vue/src/App.vue
+++ b/src-vue/src/App.vue
@@ -137,11 +137,21 @@ export default {
   font-weight: bold;
   font-size: large;
   background-color: transparent !important;
+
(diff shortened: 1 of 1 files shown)


Jan200101:
actually, for controller input I think it would be better to outright disable tab indexing the menu and instead binding that stuff to shoulder buttons.

It would inhibit keyboard users though

GeckoEidechse (approved):
Confirmed in quick testing on Windows

image

PR #967, opened 2024-07-05
fix: Add missing translation

GeckoEidechse:
Translation for showing NSFW mod was missing from #964


GeckoEidechse:
forgot to merge this, whoops

PR #969, opened 2024-07-06
feat: Add initial CGNAT check logic

GeckoEidechse:
Adds some logic that gets external IP and runs `tracert` to count the number of hops to said IP address. The goal is to have an automated way to check for CGNAT.

TODOs:

- [x] Use `-d` arg to disable reverse DNS lookups (thanks @p0358 via https://discord.com/channels/920776187884732556/936105663942246410/1262421359271743488)
- [x] Set max hope count arg
- [x] Set timeout arg
- [x] Also count lines with timouts. Currently hops are counted by checking if they contain `ms` (indicating ping time) but that skips hops that have a timeout and look like: `5   *   *   *     Request timed out.`


GeckoEidechse (commented):
Worked for me in testing and locked behind dev view for now.

PR #995, opened 2024-08-02
feat: Pick random adjective when generating release notes

GeckoEidechse:
Pick random adjective when generating release notes from a list of predefined ones


GeckoEidechse (commented):
Not much to say here. Dev feature and works for me so no further testing needed.

PR #1000, opened 2024-08-14
fix: Remove unnecessary unwraps

GeckoEidechse:
We already have `anyhow` for boxing the error so we can just bubble it up

Should fix https://northstar-kv.sentry.io/share/issue/e40587c4b0cb40038b234a17e7a0bf78/


GeckoEidechse:
Also, just noticed, we just hit 1k PRs :exploding_head:

PR #1002, opened 2024-08-29
feat: Switch location of Mods and Changelog tabs

GeckoEidechse:
Usually players care more about mods than changelogs so it should be further in front.

Old:
image

New:
image
(stripes are from dev build and unrelated)


Alystrasz (approved):
Code looks good to me, works well on Windows 10.

PR #1051, opened 2024-12-09
feat: Implement mod versioning

Alystrasz:
Implementation of https://github.com/R2Northstar/NorthstarLauncher/pull/828 new manifest format.

##### TODOs

- [ ] Fix `set_mod_enabled_status` with mods we don't know the version to enable/disable (*e.g.* core mods/mod settings)
- [ ] Add a manifest version parameter to `rebuild_enabled_mods_json`

### DIFF
--- a/src-tauri/src/mod_management/mod.rs
+++ b/src-tauri/src/mod_management/mod.rs
@@ -384,17 +384,28 @@ pub fn get_installed_mods_and_properties(
     // enabledmods.json now has a different format, forcing us to detect whether it is used and
     // react accordingly.
     //
-    // Old format: { modName: string => is_enabled: bool }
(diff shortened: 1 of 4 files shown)


Alystrasz:
@GeckoEidechse what should happen when flightcore can't load `enabledmods.json` file (thus not being able to determine manifest version being used)?
Should we rebuild `enabledmods.json` using v0 or v1?

GeckoEidechse:
I'd say rebuild in v0 for now but have a feature flag or simple conditional check to tell it to rebuild in v1 instead. That way we can just "flip the switch" when we release the corresponding update in Northstar.

GeckoEidechse:
Clippy CI errors are from update in Rust toolchain and will be fixed by #1053

PR #1091, opened 2025-04-06
Update Vanilla+ name check

NachosChipeados:
Updates Vanilla+'s name check to include the new name. Still kept the old one just in case


GeckoEidechse (approved):
Confirmed working in testing
image


Trivial code change so not much to say here. Gonna fix formatting and then merge it ^^

PR #1096, opened 2025-04-27
feat: Add initial logic for update check from frontend

### LINKED ISSUE #1095: Add updater logic
In Tauri v1 I relied on the build-in updater dialog. This was removed in Tauri v2 so we have to write our own logic

https://v2.tauri.app/start/migrate/from-tauri-1/#migrate-to-updater-plugin
Image

GeckoEidechse:
Closes #1095

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1303,6 +1303,7 @@ dependencies = [
  "tauri-build",
  "tauri-plugin-dialog",
  "tauri-plugin-opener",
+ "tauri-plugin-process",
  "tauri-plugin-store",
  "tauri-plugin-updater",
  "tokio",
@@ -5286,6 +5287,16 @@ dependencies = [
  "url",
  "windows 0.58.0",
  "zbus 4.0.1",
+]
+
+[[package]]
+name = "tauri-plugin-process"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ae06a00087c148962a52814a2d7265b1a0505bced5ffb74f8c284a5f96a4d03d"
+dependencies = [
+ "tauri",
+ "tauri-plugin",
 ]
 
 [[package]]
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -78,6 +78,7 @@ octocrab = "0.38.0"
 # Library for removing markdown links
 remove-markdown-links = "1.0.0"
 tauri-plugin-store = "2"
+tauri-plugin-process = "=2.0.1"
 
 
 [target.'cfg(windows)'.dependencies]
--- a/src-tauri/capabilities/desktop.json
+++ b/src-tauri/capabilities/desktop.json
@@ -9,6 +9,12 @@
     "main"
   ],
   "permissions": [
-    "updater:default"
+    "dialog:default",
+    "dialog:allow-ask",
+    "dialog:allow-message",
+    "updater:default",
+    "updater:allow-check",
+    "updater:allow-download-and-install",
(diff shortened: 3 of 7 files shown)


GeckoEidechse:
Okay, this is almost ready to go, I just need to figure out where to place the update checking logic in frontend. @Alystrasz any suggestions? ^^

Alystrasz:
> Okay, this is almost ready to go, I just need to figure out where to place the update checking logic in frontend. @Alystrasz any suggestions? ^^

Sorry for the late answer, thought I answered already!
I'd suggest placing it in the most top-level component, namely `App.vue`.

Alystrasz:
> > Okay, this is almost ready to go, I just need to figure out where to place the update checking logic in frontend. @Alystrasz any suggestions? ^^
> 
> Sorry for the late answer, thought I answered already! I'd suggest placing it in the most top-level component, namely `App.vue`.

Actually, we can do even better, and put it in the `store.ts` module.
@GeckoEidechse there is already a [`_checkForFlightCoreUpdates`](https://github.com/R2NorthstarTools/FlightCore/blob/main/src-vue/src/plugins/store.ts#L467) method in there, should I replace it with your new update logic?

Alystrasz:
Well, done in ed42fd25c25c470f8a3c5c0476a9f88ccbecef6d.
@GeckoEidechse how do you simulate a new flightcore version being available? (to test the popup)

GeckoEidechse:
>  @GeckoEidechse how do you simulate a new flightcore version being available? (to test the popup)

Uh, idk if Tauri does the check for `!=` or `>` current version. Need to check the docs. I think you can set it to strict and then it should prompt for update on any version difference, though idk if you also need the signing key to be set for that

src-vue/src/plugins/store.ts
```diff
@@ -479,6 +488,42 @@ async function _checkForFlightCoreUpdates(state: FlightCoreStore) {
...
+        console.log("No update available");
+    } else if (update?.available) {
+        console.log("Update available!", update.version, update.body);
+        const accepted = await ask(
+        `Update to ${update.version} is available!\n\nRelease notes: ${update.body}`,
+        {
+            title: "Update Available",
+            kind: "info",
+            okLabel: "Update",
+            cancelLabel: "Cancel",
```
GeckoEidechse, line 503: Should we translate this?

Alystrasz: At least the `title` entry, yep; I think we can reuse existing entries for the 3 other items.

src-vue/src/plugins/store.ts
```diff
@@ -479,6 +488,42 @@ async function _checkForFlightCoreUpdates(state: FlightCoreStore) {
...
 }
 
+async function flightcoreUpdateCheck() {
+    const update = await check();
+    if (!update?.available) {
+        console.log("No update available");
+    } else if (update?.available) {
+        console.log("Update available!", update.version, update.body);
+        const accepted = await ask(
+        `Update to ${update.version} is available!\n\nRelease notes: ${update.body}`,
```
Alystrasz, line 498: Actually, we should also translate this one.

GeckoEidechse:
Then we'll have to test in prod :>

src-vue/src/plugins/store.ts
GeckoEidechse: Addressed in 655f8f32b04df0a10afc708d806530863c67febb

GeckoEidechse (commented):
Seems to work in quick testing. At least it triggers the updating mechanism and launches the updater process. Whether it properly installs the update is something I cannot verify with debug build but assuming the mechanism is correctly implemented in Tauri upstream it should just work™ :)

